PR #11, opened 2021-11-29
Enhance error messages when accessing qubit and registers

idavis:
When running the generator sample it is easy to run it with too few qubits and it will crash. This PR adds two layers of error messages for API consumers to give them a better idea of what failed.

### DIFF
--- a/examples/generator/mock_to_qir.py
+++ b/examples/generator/mock_to_qir.py
@@ -27,7 +27,8 @@ class QirGenerator(MockLanguageListener):
         self.builder = QirBuilder(module_id)
         self.builder.add_quantum_register("q", nr_qubits)
         self.builder.add_classical_register("m", nr_qubits)
-        
+        self.nr_qubits = nr_qubits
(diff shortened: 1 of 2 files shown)


examples/generator/mock_to_qir.py
```diff
@@ -40,15 +41,27 @@ def write_to_file(self, file_path: str) -> str:
             file.write(self.ir_string)
 
     def enterXGate(self, ctx: MockLanguageParser.XGateContext):
+        if int(ctx.target.text) >= self.nr_qubits:
+            raise ValueError("Parsed progam uses more qubits than allocated")
```
samarsha, line 45: Maybe you could factor out this check into a function?

samarsha (approved):

examples/generator/mock_to_qir.py
idavis: @samarsha Thanks for the push. Updated it.

PR #13, opened 2021-11-29
Update Instructions for Examples

ScottCarda-MS:
Updated the installation instructions to make it more clear that the conda command needs to be run from within a conda environment.
Added a bit more instruction on how to run the examples to the example readme files.

### DIFF
--- a/examples/generator/README.md
+++ b/examples/generator/README.md
@@ -29,8 +29,8 @@ package:
   pip install antlr4-python3-runtime
   ```
 
-  The example can then be run using python, with the generated QIR being outputted
-  to a text file:
+  The example can then be run using python, with the generated QIR being
+  outputted to a text file:
 
   ```bash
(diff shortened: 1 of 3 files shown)


examples/generator/README.md
```diff
@@ -22,13 +22,20 @@ package:
...
-  and omitted any further compilation or optimization. To run the example,
+  and omitted any further compilation or optimization. Before running the example,
   please install the Antlr runtime:
 
   ```bash
   pip install antlr4-python3-runtime
   ```
 
+  The example can then be run using python, with the generated QIR being
+  outputted to a text file:
```
DmitryVasilevsky, line 33: Would 'written' sound better?

ScottCarda-MS: done

DmitryVasilevsky (approved):

bettinaheim (approved):

PR #16, opened 2021-11-30
Updating pyo3 and maturin to latest versions.

idavis:
- The 0.15.x pyo3 is the last version to support Python 3.6 which is [EOL Dec 2021](https://www.python.org/dev/peps/pep-0494/#id16)


samarsha (approved):

cgranade (approved):
It may be worth adding the terms highlighted by the spell check CI job to cspell.json so as to cut down on false positive warnings.

PR #18, opened 2021-11-30
Format rust workspace source code using cargo fmt

nilslice:
This commit was created from the running of `cargo fmt --all` from the root of the workspace. Some text editors / IDEs have a "Format on Save" option, which when enabled will execute `cargo fmt` on a per-file basis.

It might be nice to have formatting checks added to an Action at some point. I'd be happy to follow up with another PR to add such a check.


nilslice:
Re-opening rebased PR on top of `main` with recent changes.

PR #21, opened 2021-12-01
Adding cargo fmt/clippy checks on build.

idavis:


pyqir-generator/src/python.rs
```diff
@@ -30,11 +33,16 @@ impl PyQIR {
...
 
+    #[allow(clippy::needless_pass_by_value)]
+    #[allow(clippy::unnecessary_wraps)]
+    #[allow(clippy::unused_self)]
     fn add_measurement(&mut self, qubit: String, target: String) -> PyResult<()> {
         log::info!("measure {} => {}", qubit, target);
         Ok(())
     }
 
+    #[allow(clippy::needless_pass_by_value)]
```
samarsha, line 44: Is Clippy really warning here? Doesn't `Controlled::new` consume the strings?

pyqir-generator/src/python.rs
```diff
@@ -30,11 +33,16 @@ impl PyQIR {
...
+    #[allow(clippy::unnecessary_wraps)]
```
samarsha, line 45: Is it necessary to return `PyResult<()>` instead of `()`?

pyqir-jit/src/intrinsics.rs
```diff
@@ -197,11 +322,19 @@ pub unsafe extern "C" fn __quantum__qis__dumpregister__body(
     todo!("Not yet implemented.");
 }
 
+/// # Safety
+///
+/// This function should not be called directly. It is intended to be
+/// called by QIR applications during JIT execution.
 pub unsafe fn get_qubit_id(ctls: *mut QirArray) -> QUBIT {
-    let ctrl_qubit_ptr = QirRuntime::quantum_rt_array_get_element_ptr_1d(ctls, 0) as *mut u64;
+    //let ctrl_qubit_ptr = QirRuntime::quantum_rt_array_get_element_ptr_1d(ctls, 0) as *mut u64;
```
samarsha, line 330:
```suggestion
```

pyqir-jit/src/jit.rs
```diff
@@ -27,22 +31,30 @@ pub fn run_module_file(
...
-    module: &Module<'ctx>,
-    entry_point: Option<&str>,
-) -> Result<SemanticModel, String> {
-    Target::initialize_native(&InitializationConfig::default()).unwrap();
+/// # Errors
+///
+/// Will return `Err` if LLVM native target fails to initialize
+pub fn run_module(module: &Module<'_>, entry_point: Option<&str>) -> Result<SemanticModel, String> {
+    Target::initialize_native(&InitializationConfig::default())
+        .expect("Failed to initialize native target.");
```
samarsha, line 45: The comment implies that this shouldn't panic but instead return the error.

pyqir-jit/src/jit.rs
```diff
@@ -94,6 +106,7 @@ fn choose_entry_point<'ctx>(
     }
 }
 
+#[allow(clippy::trivially_copy_pass_by_ref)]
```
samarsha, line 109: I didn't realize `FunctionValue` has `Copy`. You can remove the reference here.

eng/psakefile.ps1
```diff
@@ -248,6 +262,11 @@ function Build-PyQIR([string]$project) {
             exec {
                 docker run --rm $userSpec -v $ioVolume -v $llvmVolume -e LLVM_SYS_110_PREFIX=/usr/lib/llvm -w /io/$project manylinux2014_x86_64_maturin conda run --no-capture-output python -m tox -e test
             }
+            
+            Write-BuildLog "docker run --rm $userSpec -v $ioVolume -v $llvmVolume -e LLVM_SYS_110_PREFIX=/usr/lib/llvm -w /io/$project manylinux2014_x86_64_maturin conda run --no-capture-output cargo test --release -vv -- --nocapture" "command"
+            exec {
+                docker run --rm $userSpec -v $ioVolume -v $llvmVolume -e LLVM_SYS_110_PREFIX=/usr/lib/llvm -w /io/$project manylinux2014_x86_64_maturin conda run --no-capture-output cargo test --release -vv -- --nocapture
+            }
```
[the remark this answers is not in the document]
idavis: Logging commands in general is dangerous, so I've been logging them manually.

pyqir-generator/src/python.rs
idavis: Yes, it wants `&str` as that is all the logging needs but I wanted all of the python/rust interfaces to have owned objects.

pyqir-generator/src/python.rs
idavis: When I looked originally, I thought so, we can check again.

pyqir-jit/src/jit.rs
idavis: The Errors does concern the associated message. It also can panic, but clippy didn't catch the change. Should we add a return error instead and update the doc?

pyqir-jit/src/jit.rs
idavis: Changing the ref to `FunctionValue` makes the code above it no longer work as the calls expect `&FunctionValue`.

pyqir-generator/src/interop.rs
```diff
@@ -8,10 +8,12 @@ pub struct QuantumRegister {
 }
 
 impl QuantumRegister {
+    #[must_use]
```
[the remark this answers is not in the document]
idavis: -[clippy must_use_candidate](https://rust-lang.github.io/rust-clippy/master/index.html#must_use_candidate)
-[rustlang rfcs](https://rust-lang.github.io/rfcs/1940-must-use-functions.html)

pyqir-generator/src/python.rs
idavis: @samarsha  We can remove the `PyResult<()>` for anything which doesn't return errors. I'll start on that.

idavis: @samarsha Pushed changes.

eng/psakefile.ps1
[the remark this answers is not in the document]
samarsha: I'm not referring to logging in general, but a way to avoid duplicating the command. Something like this (pseudocode, since I don't know the best way to do it in PowerShell):

```
trace(command) {
    print command as a string
    execute command
}

trace(docker --rm $userSpec ...)
```

pyqir-generator/src/python.rs
samarsha: I just tried it. Removing `#[allow(clippy::needless_pass_by_value)]` from `PyQIR::cx`, `PyQIR::cz`, etc. doesn't produce any lints. Are you thinking of `PyQIR::add_measurement`?

pyqir-jit/src/jit.rs
samarsha: You're right. You could change the `filter` call to a lambda that dereferences it, but I'm fine either way.

pyqir-jit/src/jit.rs
samarsha: I'm not sure I understand - if you change this line to a return an error, then I don't think the doc needs to be updated, because this error case falls under "return `Err` if LLVM native target fails to initialize". However, if you leave it as a panic, then I think the doc needs to be updated.

pyqir-generator/src/python.rs
```diff
@@ -212,9 +215,14 @@ impl PyQIR {
...
     fn enable_logging(&self) -> PyResult<()> {
-        let _ = env_logger::try_init();
-        Ok(())
+        match env_logger::try_init() {
+            Err(msg) => {
+                let err: PyErr = PyOSError::new_err::<String>(format!("{}", msg));
+                Err(err)
+            }
+            Ok(_) => Ok(()),
+        }
```
samarsha, line 226:
```suggestion
        env_logger::try_init().map_err(|e| PyOSError::new_err(e.to_string()))
```

pyqir-jit/src/runtime.rs
```diff
@@ -5,11 +5,13 @@ use crate::gates::CURRENT_GATES;
...
 use inkwell::execution_engine::ExecutionEngine;
 use inkwell::module::Module;
+
 use qirlib::intrinsics::Intrinsics;
 
 use super::gates::GateScope;
 
 pub(crate) struct Simulator {
+    #[allow(unused)]
     scope: GateScope,
```
samarsha, line 15:
```suggestion
    _scope: GateScope,
```

qirlib/src/context.rs
```diff
@@ -10,23 +10,26 @@ use crate::{
     types::Types,
 };
 
-pub struct BareContext<'ctx> {
+pub struct Bare<'ctx> {
```
samarsha, line 13: I think having "bare context" and "context" made sense, and you can guess the relationship between them, but not with "bare" and "context."

qirlib/src/context.rs
```diff
@@ -43,32 +46,36 @@ pub struct Context<'ctx> {
     pub constants: Constants<'ctx>,
 }
 
-pub enum ContextType<'ctx> {
+#[derive(Clone, Copy)]
+pub enum ModuleType<'ctx> {
```
samarsha, line 50: I think that this name is more accurate than `ContextType`, but I'm not sure it's quite right either. If you load a module from a template, a file, or (in the future) a byte array, the resulting module is the same type in all cases - the only difference is where it came from. Maybe it's a `ModuleLoader` or something like that?

qirlib/src/module.rs
```diff
@@ -36,21 +36,25 @@ pub(crate) fn load_module_from_bitcode_template<'ctx>(
...
 }
 
-pub(crate) fn load_module_from_bitcode_file<'ctx, P: AsRef<Path>>(
-    path: P,
-    context: &'ctx inkwell::context::Context,
-) -> Result<Module<'ctx>, String> {
+pub(crate) fn load_module_from_bitcode_file(
+    path: impl AsRef<Path>,
+    context: &'_ inkwell::context::Context,
+) -> Result<Module<'_>, String> {
```
samarsha, line 42: I think you can always omit `'_` when there's only a single lifetime parameter.

```suggestion
    context: &inkwell::context::Context,
) -> Result<Module, String> {
```

qirlib/src/module.rs
```diff
@@ -36,21 +36,25 @@ pub(crate) fn load_module_from_bitcode_template<'ctx>(
...
 }
 
-pub(crate) fn load_module_from_ir_file<'ctx, P: AsRef<Path>>(
-    path: P,
-    context: &'ctx inkwell::context::Context,
-) -> Result<Module<'ctx>, String> {
+pub(crate) fn load_module_from_ir_file(
+    path: impl AsRef<Path>,
+    context: &'_ inkwell::context::Context,
+) -> Result<Module<'_>, String> {
```
samarsha, line 49:
```suggestion
    context: &inkwell::context::Context,
) -> Result<Module, String> {
```

qirlib/src/passes.rs
```diff
@@ -3,13 +3,10 @@
...
 use inkwell::passes::PassManager;
-use inkwell::{
-    passes::PassManagerBuilder,
-    OptimizationLevel,
-};
+use inkwell::{passes::PassManagerBuilder, OptimizationLevel};
 
 // This method returns true if any of the passes modified the function or module and false otherwise.
-pub fn run_basic_passes_on<'ctx>(module: &Module<'ctx>) -> bool {
+pub fn run_basic_passes_on(module: &Module<'_>) -> bool {
```
samarsha, line 9:
```suggestion
pub fn run_basic_passes_on(module: &Module) -> bool {
```

pyqir-generator/src/python.rs
idavis: You are right. Thank you!

eng/psakefile.ps1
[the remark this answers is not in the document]
idavis: @samarsha Pushed new Invoke-LoggedCommand. Please take a look.

qirlib/src/context.rs
idavis: I agree.
- `ModuleLoader`
- `ModuleSource`
- `ModuleType`
- ?

pyqir-generator/src/python.rs
```diff
@@ -30,139 +30,125 @@ impl PyQIR {
         }
     }
 
-    fn add_measurement(&mut self, qubit: String, target: String) -> PyResult<()> {
+    #[allow(clippy::needless_pass_by_value)]
```
cgranade, line 32: It may be helpful to have a brief comment explaining these two attributes and why they're useful here.

pyqir-generator/src/python.rs
```diff
@@ -30,139 +30,125 @@ impl PyQIR {
...
+    #[allow(clippy::unused_self)]
+    fn add_measurement(&mut self, qubit: String, target: String) {
```
cgranade, line 34: Will it be a breaking change to the API if you need to return an Err in the future?

pyqir-generator/src/qir/array1d.rs
```diff
@@ -96,30 +96,6 @@ fn get_bitcast_qubit_pointer_element<'ctx>(
     cast
 }
 
-fn get_bitcast_array_element<'ctx>(
```
cgranade, line 99: Why was this function removed, out of curiousity?

pyqir-generator/src/qir/array1d.rs
```diff
@@ -191,15 +166,15 @@ pub(crate) fn emit_array_get_element_ptr_1d<'ctx>(
...
     context: &Context<'ctx>,
     results: &BasicValueEnum<'ctx>,
-    sub_results: Vec<BasicValueEnum<'ctx>>,
+    sub_results: &[BasicValueEnum<'ctx>],
     name: &str,
-) -> () {
-    for index in 0..sub_results.len() {
-        let result_indexed_name = format!("{}_result_tmp", &name[..]);
+) {
+    for (index, _) in sub_results.iter().enumerate() {
```
cgranade, line 172: Ah, that's a cute way of getting the index range!

pyqir-jit/src/gates.rs
```diff
@@ -56,22 +68,23 @@ impl BaseProfile {
     fn record_max_qubit_id(&mut self, qubit: QUBIT) {
         self.declared_cubits = true;
         if qubit > self.max_id {
-            self.max_id = qubit
+            self.max_id = qubit;
         }
     }
+    #[must_use]
```
cgranade, line 74: Formatting nit, but is a blank line required here?

pyqir-jit/src/intrinsics.rs
```diff
@@ -21,13 +21,24 @@ fn get_current_gate_processor() -> ForceSomeRwLockWriteGuard<'static, BaseProfil
...
     log::debug!("/__quantum__qis__h__body/");
     let mut gs = get_current_gate_processor();
     gs.h(qubit);
 }
 
+/// # Safety
+///
+/// This function should not be called directly. It is intended to be
+/// called by QIR applications during JIT execution.
+/// # Panics
```
cgranade, line 39: Nit, but Markdown headers should be separated by at least one preceding blank line.

pyqir-jit/src/jit.rs
```diff
@@ -27,22 +31,30 @@ pub fn run_module_file(
...
+        .expect("Failed to initialize native target.");
+
     let default_triple = TargetMachine::get_default_triple();
     let target = Target::from_triple(&default_triple).expect("Unable to create target machine");
-    assert!(target.has_asm_backend());
-    assert!(target.has_target_machine());
 
-    run_basic_passes_on(&module);
+    if !target.has_asm_backend() {
+        return Err("Target doesn't have an ASM backend.".to_owned());
```
cgranade, line 68: It looks like these Err cases may not be included in the `# Errors` section above?

qirlib/src/context.rs
idavis: I agree. We need a better name. Clippy complained about the `Context` suffix. We can change it back but we'll have to attribute it.

eng/psakefile.ps1
```diff
@@ -25,7 +25,21 @@ properties {
...
+
+Task cargo-fmt {
+    Invoke-LoggedCommand -workingDirectory $repo.root -errorMessage "Please run 'cargo fmt --all' before pushing" {
+        cargo fmt --all -- --check
+    }
+}
+
+Task cargo-clippy -Depends init {
+    Invoke-LoggedCommand -workingDirectory $repo.root -errorMessage "Please fix the above clippy errors" {
+        cargo clippy --workspace --all-targets --all-features -- -D warnings
```
samarsha, line 40: It seems like any Clippy lints are treated as errors when building locally. I am using Rust 1.57 and Clippy apparently has an additional lint, and the build fails for me when I run `build.ps1`. IMO, it's better to keep warnings as warnings locally, and only elevate them to errors in CI. That means that minor issues during development can be deferred to later without requiring immediate attention, but the code merged to main is still warning-clean.

samarsha: Specifically, in the packages' lib.rs files: https://github.com/qir-alliance/pyqir/blob/19df229766d6b3dc0efc986e7a030a30bc88461a/qirlib/src/lib.rs#L4

I think this `deny` should be changed to `warn`, and the scripts should set it back to deny with a command-line option only during CI.

qirlib/src/context.rs
samarsha: I'd be ok with `ModuleSource` or `ModuleLoader`. I can't think of any other names right now.

qirlib/src/codegen.rs
```diff
@@ -1,111 +1,64 @@
...
-    pub builder: inkwell::builder::Builder<'ctx>,
     pub types: Types<'ctx>,
     pub runtime_library: RuntimeLibrary<'ctx>,
     pub intrinsics: Intrinsics<'ctx>,
     pub constants: Constants<'ctx>,
 }
 
-pub enum ContextType<'ctx> {
+#[derive(Clone, Copy)]
+pub enum ModuleSource<'ctx> {
```
samarsha, line 22: Should ModuleSource be moved to qirlib::module?

idavis: @samarsha Done.

qirlib/src/module.rs
```diff
@@ -5,29 +5,38 @@ use inkwell::{memory_buffer::MemoryBuffer, module::Module};
 
 use std::path::Path;
 
-use crate::context::ContextType;
+#[derive(Clone, Copy)]
+pub enum ModuleSource<'ctx> {
```
samarsha, line 9: Name fails clippy::module-name-repetitions lint. I think I agree with Clippy, module::Source seems better than module::ModuleSource.

idavis: @samarsha Updating.

samarsha (approved):

PR #23, opened 2021-12-01
Support running modules with a custom entry point name or multiple entry points

samarsha:


pyqir-jit/src/jit.rs
```diff
@@ -24,56 +28,187 @@ pub fn run_module_file<P: AsRef<Path>>(path: P) -> Result<SemanticModel, String>
...
         let _ = microsoft_quantum_qir_runtime_sys::foundation::QSharpFoundation::new();
-
         let _ = inkwell::support::load_library_permanently("");
         let execution_engine = module
             .create_jit_execution_engine(OptimizationLevel::None)
             .expect("Could not create JIT Engine");
         let simulator = Simulator::new(&module, &execution_engine);
         let main = execution_engine
-            .get_function::<unsafe extern "C" fn() -> ()>("QuantumApplication__Run")
+            .get_function::<unsafe extern "C" fn() -> ()>(&entry_point)
```
idavis, line 57: @samarsha Is this still a safe type?

pyqir-jit/src/jit.rs
```diff
@@ -24,56 +28,187 @@ pub fn run_module_file<P: AsRef<Path>>(path: P) -> Result<SemanticModel, String>
...
     }
 }
 
+fn choose_entry_point<'ctx>(
+    functions: impl Iterator<Item = FunctionValue<'ctx>>,
+    name: Option<&str>,
+) -> Result<String, String> {
+    let mut entry_points = functions
+        .filter(is_entry_point)
+        .map(|f| f.get_name().to_str().unwrap().to_owned())
```
idavis, line 71: Should we use expect here instead of unwap?

pyqir-jit/src/jit.rs
samarsha: I added a check before running the entry point to make sure it has no parameters and a void return type.

pyqir-jit/src/jit.rs
samarsha: I refactored this: instead of unwrapping the function name, I wrap the provided entry point name in `Ok` when comparing. If converting the CStr to a str fails, then the names can't match so we can ignore it.

idavis (approved):

PR #24, opened 2021-12-04
Support loading modules from memory and clean up module loading API

samarsha:
Added support for loading QIR modules directly from a byte array, avoiding the need to create temporary files in the JIT tests.

I removed the `module::Source` type to make the module loading API more compositional. Instead of passing a `Source` enum to a function that needs a module, you pass the module in directly, so the function doesn't need to know how to load the module (increasing separation of concerns).

`qirlib::module` provides some loading functions that correspond to the original `Source` cases, but unlike an enum, it doesn't imply that they are exhaustive.


pyqir-jit/src/python.rs
```diff
@@ -73,7 +76,9 @@ impl PyNonadaptiveJit {
             Ok(())
         }
 
-        let gen_model = jit::run_module_file(file, entry_point).map_err(PyOSError::new_err)?;
+        let context = Context::create();
```
idavis, line 79: I would prefer to put the `run_module_file` method back in. I don't think we should be surfacing inkwell and qirlib in the python interop layer.

idavis (changes requested):

idavis (approved):

PR #37, opened 2022-01-27
Fix documentation links that couldn't be checked as part of the first docs PR

idavis:
The link checks for the original docs PR somehow passed with broken links. We still have the issue that we can't point to docs until after publishing, but we'll need to address that in a separate PR.

### DIFF
--- a/README.md
+++ b/README.md
@@ -36,10 +36,10 @@ by the above Python packages.
 ## Documentation
 
 - [Installing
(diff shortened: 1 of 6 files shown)


cgranade (approved):
LGTM! It may be good to add "pyqir" to the allow list to cut down on some of the false positives, thankfully those are just warnings.

PR #41, opened 2022-02-02
Cache properties as member variables

### LINKED ISSUE #39: QirBlock does not have a stable hash
In the definition for the `QirBlock` class, when you ask for the list of instructions it constructs a new list each time. This makes the hash unstable then when iterating through the functions of a block at different times. It seems like the `QirBlock` should either be unhashable or have it pass on the hash from the block. 

https://github.com/qir-alliance/pyqir/blob/c68d735331158f655db965daae8d925b8a1b975c/pyqir-parser/pyqir_parser/parser.py#L723

swernli:
Fixes #39

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -2,6 +2,8 @@
 
 ## [Unreleased]
 
+- Cache parser values for better hashing in Python [#41](https://github.com/qir-alliance/pyqir/pull/41)
+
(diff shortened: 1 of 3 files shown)


idavis (approved):

swernli:
I need to update the changelog before merging.

Issue #29, opened 2022-01-04
pyqir_parser exposing too many things on import


crazy4pi314:
I was working with the parser package, and saw that when you import it, it lists a lot of things that probably shouldn't.
Things like `List` and `Optional`, the underlying Rust interface (things prefixed by "Py"), and the `parser`/`pyqir_parser` also related to internals that probably shouldn't be user facing. 

The example below was generated from using this [conda env](https://github.com/crazy4pi314/qwop/blob/master/environment.yaml), and locally downloading and installing the windows wheels from the [v0.1.0a1 pre-release](https://github.com/qir-alliance/pyqir/releases/tag/v0.1.0a1) on GH.

```
import pyqir_parser as pqp
dir(pqp)

['List',
 'Optional',
 'PyQirBasicBlock',
 'PyQirConstant',
 'PyQirFunction',
 'PyQirInstruction',
 'PyQirModule',
 'PyQirOperand',
 'PyQirParameter',
 'PyQirTerminator',
 'PyQirType',
 'QirAShrInstr',
 'QirAddInstr',
 'QirAndInstr',
 'QirArrayType',
 'QirBlock',
 'QirBrTerminator',
 'QirCallInstr',
 'QirCondBrTerminator',
 'QirConstant',
 'QirDoubleConstant',
 'QirDoubleType',
 'QirFAddInstr',
 'QirFCmpInstr',
 'QirFDivInstr',
 'QirFMulInstr',
 'QirFNegInstr',
 'QirFRemInstr',
 'QirFSubInstr',
 'QirFunction',
 'QirICmpInstr',
 'QirInstr',
 'QirIntConstant',
 'QirIntegerType',
 'QirLShrInstr',
 'QirLocalOperand',
 'QirModule',
 'QirMulInstr',
 'QirNamedStructType',
 'QirNullConstant',
 'QirOpInstr',
 'QirOperand',
 'QirOrInstr',
 'QirParameter',
 'QirPhiInstr',
 'QirPointerType',
 'QirQirCallInstr',
 'QirQisCallInstr',
 'QirQubitConstant',
 'QirQubitType',
 'QirResultConstant',
 'QirResultType',
 'QirRetTerminator',
 'QirRtCallInstr',
 'QirSDivInstr',
 'QirSRemInstr',
 'QirShlInstr',
 'QirStructType',
 'QirSubInstr',
 'QirSwitchTerminator',
 'QirTerminator',
 'QirType',
 'QirUDivInstr',
[...]

PR #43, opened 2022-02-03
PyQIR generator API redesign with classical if statement

### LINKED ISSUE #32: PyQIR generator API redesign
The PyQIR generator API needs to be redesigned to support more QIR/LLVM features, like #2. A draft of the new API is being worked on here (comments welcome): https://gist.github.com/samarsha/23ee1e3178f12d387c08429a99bb1e4b

- [x] Implement Python native module for new API
- [x] Polish exposed Python API (re-exports, Mypy stub files, doc strings, etc.)
- [x] Update tests

samarsha:
Closes #32.


pyqir-generator/pyqir/generator/__init__.py
```diff
@@ -0,0 +1,2 @@
+# Copyright (c) Microsoft Corporation.
+# Licensed under the MIT License.
```
swernli, line 2: For my edification, what's the role of the empty init file here? Is that a common pattern in python development?

pyqir-generator/pyqir/generator/qis.pyi
```diff
@@ -0,0 +1,169 @@
...
+        one: Callable[[], None] = ...,
+        zero: Callable[[], None] = ...,
+    ) -> None:
+        """
+        Builds a conditional branch on the result of a measurement.
+
+        Dereferences the result reference, then evaluates the instructions
+        built by ``one`` if the result is one, or the instructions built
+        by ``zero`` if the result is zero. The one and zero callables should use
+        this builder to build instructions.
```
swernli, line 161: Regarding the need to use the same builder within the callables, is there any way to enforce that? Just thinking out loud, could it be part of the convention that the builder passes itself into the callables? Nothing for this PR, just food for thought.

pyqir-generator/pyqir/generator/value.pyi
```diff
@@ -0,0 +1,11 @@
+# Copyright (c) Microsoft Corporation.
+# Licensed under the MIT License.
+
+class Qubit:
+    """A qubit."""
```
swernli, line 5: Nitpick: Is it a qubit or is it a qubit identifier?

pyqir-generator/src/python.rs
```diff
@@ -1,174 +1,264 @@
...
+            instructions: Vec::new(),
+            static_alloc: true,
+        };
+
+        let builder = Py::new(py, Builder::new())?;
+        Ok(SimpleModule { model, builder })
+    }
+
+    #[getter]
+    fn qubits(&self) -> Vec<Qubit> {
```
swernli, line 54: This getter seems to create a new vector each time it's called, which can result in the items at the python level having different hashes. This may be less of a concern for generation than for parsing, since in the latter case the QIR is constant while the former is necessarily mutating it, but it's something I figured I'd call out since it came from a community issue (see https://github.com/qir-alliance/pyqir/issues/39).

pyqir-generator/src/qir/instructions.rs
```diff
@@ -126,5 +126,45 @@ pub(crate) fn emit<'ctx>(
...
+
+fn emit_if<'ctx>(
+    generator: &CodeGenerator<'ctx>,
+    registers: &mut HashMap<String, Option<PointerValue<'ctx>>>,
+    qubits: &HashMap<String, BasicValueEnum<'ctx>>,
+    entry_point: FunctionValue,
+    if_inst: &If,
+) {
+    // panic if we reference a register that hasn't been declared
+    // default the register value to zero if if hasn't been measured/declared
```
swernli, line 141: This comment may be out of date, in that we do panic if the register hasn't been declared but default to zero if it hasn't been measure (remove the "/declared").

pyqir-generator/pyqir/generator/qis.pyi
swernli: I see in the Rust code that it's already setup to use the same generator. Is this comment still necessary? Is it possible from python to pass a different builder such that this expectation gets violated?

pyqir-jit/src/python.rs
```diff
@@ -2,10 +2,15 @@
...
+use bitvec::prelude::BitVec;
+use pyo3::{
+    exceptions::PyOSError,
+    prelude::*,
+    types::{PyDict, PyList},
+};
 
 #[pymodule]
-fn pyqir_jit(_py: Python<'_>, m: &PyModule) -> PyResult<()> {
+fn _native(_py: Python<'_>, m: &PyModule) -> PyResult<()> {
```
swernli, line 13: Is the `_` here somehow because the python interop doesn't let rust see that this is actually used, or is this function really not used anywhere? If the latter, can it just be removed?

pyqir-parser/Cargo.toml
```diff
@@ -24,6 +24,7 @@ name = "pyqir_parser"
 crate-type = ["cdylib"]
 
 [package.metadata.maturin]
+name = "pyqir.parser._native"
```
swernli, line 27: Ah, I see it's part of something maturin related here. Maybe a comment in the rust indicating that `_native` is used this way?

pyqir-tests/tests/test_teleport.py
```diff
@@ -0,0 +1,143 @@
...
+    qis.cx(register, target)
+
+    # Encode the message into the entangled pair.
+    qis.cx(msg, register)
+    qis.h(msg)
+
+    # Measure the qubits to extract the classical data we need to
+    # decode the message by applying the corrections on
+    # the target qubit accordingly.
+    # We use MResetZ from the Microsoft.Quantum.Measurement namespace
```
swernli, line 28: You could remove the mention of `MResetZ` and the namespace from here.

pyqir-tests/tests/test_teleport.py
```diff
@@ -0,0 +1,143 @@
...
+    # to reset our qubits as we go.
+    qis.m(msg, results[0])
+    qis.reset(msg)
+    qis.if_result(results[0], one=lambda: qis.z(target))
+
+    # We can also use library functions such as IsResultOne to write
+    # out correction steps. This is especially helpful when composing
+    # conditionals with other functions and operations, or with partial
+    # application.
```
swernli, line 37: These lines too are specific to the Q# style invocation, and not really to this version.

swernli (approved):
A few minor comments, but overall it all looks good!

eng/psakefile.ps1
```diff
@@ -68,6 +68,13 @@ Task parser -Depends init {
     Build-PyQIR($pyqir.parser.name)
 }
 
+Task pyqir-tests -Depends init {
+    Invoke-LoggedCommand -workingDirectory (Join-Path $repo.root pyqir-tests) {
+        & $python -m pip install tox
```
cgranade, line 73: It looks like this runs in whatever virtualenv was active when `psakefile.ps1` was run. It may be worth creating a new virtualenv before installing `tox` to make sure to avoid making global changes to local dev environments.

examples/jit/bernstein_vazirani.py
```diff
@@ -3,7 +3,8 @@
 # Copyright(c) Microsoft Corporation.
```
cgranade, line 3:
```suggestion
# Copyright (c) Microsoft Corporation.
```

pyqir-generator/pyqir/generator/__init__.py
cgranade: This marks the `pyqir/generator` folder as a package; otherwise, the folder is effectively invisible to Python.

Regardless, though, it might make sense to re-export items from submodules of `pyqir.generator` here so that `import pyqir.generator` is useful without necessarily needing to remember which parts are in `module` or in `qis`.

pyqir-generator/pyqir/generator/builder.pyi
```diff
@@ -0,0 +1,6 @@
+# Copyright (c) Microsoft Corporation.
+# Licensed under the MIT License.
+
+class Builder:
+    """An instruction builder."""
+    ...
```
cgranade, line 6: Do any other attributes of `Builder` need to be listed here for type checking to succeed?

pyqir-jit/pyqir/jit/nonadaptivejit.py
```diff
@@ -16,13 +16,18 @@ class NonadaptiveJit(object):
     def __init__(self):
         self.jit = PyNonadaptiveJit()
 
-    def eval(self, file_path: str, gateset: GateSet, entry_point: Optional[str] = None):
+    def eval(self,
+             file_path: str,
+             gateset: GateSet,
+             entry_point: Optional[str] = None,
+             result_stream: Optional[List[bool]] = None):
```
samarsha, line 23: Should `result_stream` be required? It may be less surprising for pyqir-jit to panic if a measurement is made past the end of the result stream, rather than silently return zero.

pyqir-generator/pyqir/generator/__init__.py
samarsha: [Every Python package must contain an `__init__.py` file.](https://docs.python.org/3/tutorial/modules.html#packages) If there's nothing to do to initialize the package, it can be empty.

pyqir-generator/pyqir/generator/qis.pyi
samarsha: > Regarding the need to use the same builder within the callables, is there any way to enforce that?

Not easily, I don't think. You could have any number of modules and builders in scope and accidentally build on the wrong one.

For example:

```python
mod1 = SimpleModule(...)
mod2 = SimpleModule(...)
qis1 = BasicQisBuilder(mod1.builder)
qis2 = BasicQisBuilder(mod2.builder)

# Oops, wrong builder in one branch.
qis1.if_result(mod1.results[0], lambda: qis2.x(mod2.qubits[0]))
```

> Just thinking out loud, could it be part of the convention that the builder passes itself into the callables?

Yes, you could have the convention that the branch generators are of the form `lambda qis: qis.x(mod.qubits[0])` or `lambda builder: BasicQisBuilder(builder).x(mod.qubits[0])` instead of a closure over the builder. However, this would only be a hint, not a requirement that the given builder should be used.

pyqir-generator/src/python.rs
```diff
@@ -1,174 +1,264 @@
...
     #[new]
-    fn new(name: String) -> Self {
-        PyQIR {
-            model: SemanticModel::new(name),
+    fn new(py: Python, name: String, num_qubits: u64, num_results: u64) -> PyResult<SimpleModule> {
+        let registers = vec![ClassicalRegister::new(RESULT_NAME.to_string(), num_results)];
+
+        let qubits = (0..num_qubits)
+            .map(|i| QuantumRegister::new(QUBIT_NAME.to_string(), i))
+            .collect();
```
cgranade, line 39: Out of curiosity, it looks like these two definitions are fairly asymmetrical — is there a reason `registers` is defined as containing a single 𝑛-bit classical register while `qubits` contains 𝑛 1-qubit registers?

pyqir-generator/src/python.rs
```diff
@@ -1,174 +1,264 @@
...
+            .map(|index| Ref(RefKind::Result { index }))
+            .collect()
+    }
+
+    #[getter]
+    fn builder(&self) -> Py<Builder> {
+        self.builder.clone()
+    }
+
+    fn ir(&mut self, py: Python) -> PyResult<String> {
```
cgranade, line 75: Why does this need `&mut self`?

pyqir-generator/src/python.rs
```diff
@@ -1,174 +1,264 @@
...
     }
+}
+
+#[pyclass]
+struct Qubit {
+    index: u64,
+}
+
+impl Qubit {
+    fn id(&self) -> String {
```
cgranade, line 103: Should this be exposed as a `__repr__` method?

pyqir-generator/src/python.rs
```diff
@@ -1,174 +1,264 @@
...
+        format!("{}{}", QUBIT_NAME, self.index)
+    }
+}
+
+#[pyclass]
+struct Ref(RefKind);
+
+impl Ref {
+    fn id(&self) -> String {
```
cgranade, line 112: Similarly here.

pyqir-generator/tests/test_api.py
```diff
@@ -1,134 +1,95 @@
...
+def test_bell() -> None:
+    module = SimpleModule("Bell circuit", num_qubits=2, num_results=2)
+    qis = BasicQisBuilder(module.builder)
+    qis.h(module.qubits[0])
+    qis.cx(module.qubits[0], module.qubits[1])
+    qis.m(module.qubits[0], module.results[0])
+    qis.m(module.qubits[1], module.results[1])
+
+    ir = module.ir()
+    assert ir.startswith("; ModuleID = 'Bell circuit'")
```
cgranade, line 17: It looks like the tests in this file only check that the IR starts with a prescribed comment; is this a good place to check other properties of output IR as well?

pyqir-jit/pyqir/jit/nonadaptivejit.py
```diff
@@ -2,8 +2,8 @@
 # Licensed under the MIT License.
 
 from typing import Optional
-from pyqir_jit import GateSet
-from .pyqir_jit import PyNonadaptiveJit
+from pyqir.jit.gateset import GateSet
+from pyqir.jit._native import PyNonadaptiveJit as PyNonadaptiveJit
```
cgranade: It looks like the `as` clause may not be needed?

pyqir-jit/src/python.rs
cgranade: If I understand correctly, this is to mark it as private to the Python module once it's imported there. That helps address issues like #29, where Rust implementation details can inadvertently become visible to pure-Python callers.

cgranade (commented):
Leaving as comments only since it's the PR is still in draft, but had a few thoughts here and there. Really exciting to see this all come together! ♥

eng/psakefile.ps1
idavis: @cgranade The pyqir-tests folder defines a tox env which they are installed into.

pyqir-generator/tests/test_api.py
idavis: These are python smoke tests to make sure it is going through correctly. The IR tests are in the Rust side.

pyqir-jit/pyqir/jit/nonadaptivejit.py
samarsha: It's been removed.

pyqir-generator/pyqir/generator/builder.pyi
samarsha: No, for the initial version of the API there are no `Builder` attributes that are exposed in Python. They are internal to Rust only.

pyqir-generator/src/python.rs
samarsha: I implemented `__hash__` and `__richcmp__` for `Qubit` and `Ref`. The hash is now based on the index they represent instead of their object identity and should be consistent across property accesses.

eng/psakefile.ps1
cgranade: Right, understood, my question was about tox itself as it looks like that task doesn't activate an environment before installing tox.

pyqir-generator/tests/test_api.py
cgranade: Makes sense, but it may be helpful to have some comments to that effect, then?

pyqir-jit/pyqir/jit/nonadaptivejit.py
samarsha: Updated the documentation for `NonadaptiveJit` to explicitly state this behavior.

pyqir-generator/src/python.rs
samarsha: This particular string should not be exposed - it's an implementation detail that may be removed in the future (the `SemanticModel` doesn't really need to use string identifiers). However, I implemented a different repr for Qubit and Ref:

```
<Qubit 0>
<Ref to Result 0>
```

pyqir-jit/src/python.rs
samarsha: It is a requirement that the name of the pymodule init function match the name of the pyd native module, which is `_native.py`. By default, `#[pymodule]` uses the name of the Rust function it's attached to for the init function.

pyqir-parser/Cargo.toml
samarsha: Is it the name needing to be the same that's confusing, or specifically the underscore? If it's the underscore, it can alternatively be defined as

```rust
#[pymodule(name = "_native")]
fn native(_py: Python<'_>, m: &PyModule) -> PyResult<()> { ... }
```

Do you prefer that?

swernli: The name confuses me just because prepending with underscore usually means a rust function or variable unused. I've seen clippy flag any unused variables that aren't prepended and flag any variables prepended that are actually used, so seeing this `_native` made me think it was in fact an unused function. But given that this is one of the places where two conventions from different languages are meeting, I am fine if the underscore is kept. Whatever makes the most sense!

pyqir-generator/pyqir/__init__.py
```diff
@@ -0,0 +1 @@
+__path__ = __import__("pkgutil").extend_path(__path__, __name__)
```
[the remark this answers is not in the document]
samarsha: Yes, Python doesn't need it, but Maturin does: PyO3/maturin#811

pyqir-parser/Cargo.toml
samarsha: I agree that it makes it look like an unused function in Rust. I renamed them to `native_module` with a pyo3 attribute to name them `_native` in the pyd.

pyqir-generator/pyqir/generator/__init__.py
samarsha: > Regardless, though, it might make sense to re-export items from submodules of `pyqir.generator` here so that `import pyqir.generator` is useful without necessarily needing to remember which parts are in `module` or in `qis`.

@cgranade I think that if we made everything available in `pyqir.generator` module, I would want to hide/remove the submodules, so that there's only one way to import them. Does that make sense?

pyqir-generator/src/python.rs
swernli: Fantastic, thanks!

pyqir-parser/Cargo.toml
swernli: Great,, thanks!

pyqir-generator/src/python.rs
samarsha: @idavis defined the register types and would know more.

I agree with you that QuantumRegister and ClassicalRegister should be symmetric. There is also the issue of the instructions in the interop module using string identifiers, which requires the new python module to unnecessarily allocate strings. I would like to clean up these issues in a separate PR.

samarsha: #53

pyqir-generator/pyqir/__init__.py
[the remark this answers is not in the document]
cgranade: Makes sense, thanks for the context! It may be helpful to add that as a comment, then, perhaps?

eng/psakefile.ps1
samarsha: This problem existed before this PR, so I created #55 to follow up on it.

idavis (approved):

pyqir-generator/pyqir/generator/__init__.py
samarsha: Created #56 to follow up on this.

pyqir-generator/src/python.rs
cgranade: Makes sense! As a quick thought should that optimization become necessary again in the future, if the mutation is not externally visible, that may be an appropriate place to use an `Rc<RefCell<_>>` to allow interior mutability without propagating to an observable mutation.

PR #46, opened 2022-02-04
Moving to minimal API

idavis:
- Removing string and arrays from default
- Adding base for static allocation and register initialization.

Dynamic  allocation:
```llvm
; ModuleID = 'Bell circuit'
source_filename = "./module.ll"

%Qubit = type opaque
%Result = type opaque

define void @QuantumApplication__Run__body() #0 {
entry:
  %qr0 = call %Qubit* @__quantum__rt__qubit_allocate()
  %qr1 = call %Qubit* @__quantum__rt__qubit_allocate()
  call void @__quantum__qis__h__body(%Qubit* %qr0)
  call void @__quantum__qis__cnot__body(%Qubit* %qr0, %Qubit* %qr1)
  %qc0 = call %Result* @__quantum__qis__m__body(%Qubit* %qr0)
  %qc1 = call %Result* @__quantum__qis__m__body(%Qubit* %qr1)
  call void @__quantum__rt__qubit_release(%Qubit* %qr1)
  call void @__quantum__rt__qubit_release(%Qubit* %qr0)
  ret void
}

declare void @__quantum__qis__cnot__body(%Qubit*, %Qubit*)

declare void @__quantum__qis__h__body(%Qubit*)

declare %Result* @__quantum__qis__m__body(%Qubit*)

declare %Qubit* @__quantum__rt__qubit_allocate()

declare void @__quantum__rt__qubit_release(%Qubit*)

attributes #0 = { "EntryPoint" }

```

Static allocation:
```llvm
; ModuleID = 'Bell circuit'
source_filename = "./module.ll"

%Qubit = type opaque
%Result = type opaque

define void @QuantumApplication__Run__body() #0 {
entry:
  call void @__quantum__qis__h__body(%Qubit* null)
  call void @__quantum__qis__cnot__body(%Qubit* null, %Qubit* inttoptr (i64 1 to %Qubit*))
  %qc0 = call %Result* @__quantum__qis__m__body(%Qubit* null)
  %qc1 = call %Result* @__quantum__qis__m__body(%Qubit* inttoptr (i64 1 to %Qubit*))
  ret void
}

declare void @__quantum__qis__cnot__body(%Qubit*, %Qubit*)

declare void @__quantum__qis__h__body(%Qubit*)

declare %Result* @__quantum__qis__m__body(%Qubit*)

attributes #0 = { "EntryPoint" "requiredQubits"="2" }
```


qirlib/src/codegen/qis.rs
```diff
@@ -0,0 +1,227 @@
...
+    module: &Module<'ctx>,
+    name: &str,
+) -> FunctionValue<'ctx> {
+    let function_name = format!("__quantum__qis__{}__body", name.to_lowercase());
+    if let Some(function) = get_function(module, function_name.as_str()) {
+        function
+    } else {
+        let void_type = context.void_type();
+        let qubit_ptr_type = qubit(context, module).ptr_type(AddressSpace::Generic);
+        let fn_type = void_type.fn_type(&[qubit_ptr_type.into()], false);
```
swernli, line 167: All of these void-typed intrinsic function getters follow very similar patterns where the only difference is the argument types. Could they be refactored to call a common base? The callers would look something like:
```rust
pub(crate) fn get_rotated_intrinsic_function_body<'ctx>(
    context: &'ctx inkwell::context::Context,
    module: &Module<'ctx>,
    name: &str,
) -> FunctionValue<'ctx> {
    let qubit_ptr_type = qubit(context, module).ptr_type(AddressSpace::Generic);
    get_intrinsic_function_body_helper(context, module, name, &[double(context).into(), qubit_ptr_type.into()])
}
```
That would mean the slice of parameter types gets constructed every time when it might not need to be, but perhaps that's a reasonable trade off to consider.

qirlib/src/codegen/types.rs
```diff
@@ -0,0 +1,122 @@
...
+#[must_use]
+pub(crate) fn result<'ctx>(
+    context: &'ctx inkwell::context::Context,
+    module: &inkwell::module::Module<'ctx>,
+) -> StructType<'ctx> {
+    get_or_define_struct(context, module, "Result")
+}
+
+#[must_use]
+pub(crate) fn array<'ctx>(
```
swernli, line 50: This one should no longer be needed, since arrays aren't used anywhere anymore.

qirlib/src/codegen/types.rs
```diff
@@ -0,0 +1,122 @@
...
+    fn result_can_be_declared() {
+        let context = Context::create();
+        let module = context.create_module("test");
+        let generator = CodeGenerator::new(&context, module).unwrap();
+
+        verify_opaque_struct("Result", result(generator.context, &generator.module));
+    }
+
+    #[test]
+    fn array_can_be_declared() {
```
swernli, line 109: Likewise this can be removed too.

swernli (commented):
I had a couple feedback items, nothing major. Really excited to see how much this approach simplifies things and removes the need for the runtime and template!

swernli (approved):

PR #79, opened 2022-03-10
Wheel/Project Renames

### LINKED ISSUE #59: Rename qir_jit to qir_evaluation
Some discussions around comprehensive naming of the packages have come up. Something like qir_evaluation rather than qir_jit may be more approachable.

idavis:
Renaming wheels in preparation for PyPI publishing
- pyqir_generator => pyqir-generator
- pyqir_parser => pyqir_parser
- pyqir_jit => pyqir-evaluator (closes #59)
  - Do we have a better name for `NonadaptiveJit` and `PyNonadaptiveJit`?
 
/CC @guenp @cgranade @samarsha @swernli


guenp (approved):

idavis:
@guenp @cgranade @samarsha @swernli

Do we have a better name for NonadaptiveJit and PyNonadaptiveJit? Or should we leave it?

swernli:
For `NonadaptiveJit`, since that is a component that a consumer of the tool would actually interact with in their code, I think we should change it to `NonadaptiveEvaluator`. But for the `PyNonadaptiveJit` my understanding is that this is an internal part of the code. If that's the case, then I'm fine leaving that one as descriptive for the developer working on pyqir, while the more generic term is used for the developing building on top of pyqir.

PR #94, opened 2022-03-29
Add licensing checks and notice file

idavis:
- CI pipeline to verify license compliance
- build task to check license compliance
- build task to generate notice file
- config files and templates for notice file
- generated notice file


eng/psakefile.ps1
```diff
@@ -427,3 +427,27 @@ function Create-DocsEnv() {
...
+task update-noticefile {
+    # use cargo-about to generate a notice file
+    # llvm special license is already in the template
+    # as it is a hidden transitive dependency.
+    # https://github.com/EmbarkStudios/cargo-about
+    $config = Join-Path $repo.root notice.toml
+    $template = Join-Path $repo.root notice.hbs
+    $notice = Join-Path $repo.root NOTICE.txt
+    Invoke-LoggedCommand -wd $repo.root {
+        cargo about generate --config $config --workspace --all-features --output-file $notice $template
```
swernli, line 449: Does the `cargo-about` tool need to get installed separately? I don't see a place where that happens, but maybe I'm missing it.

idavis: This is a manual process so you'd have to install it yourself. I don't know of a way to ask cargo if a subcommand exists.

swernli (approved):

eng/psakefile.ps1
swernli: I think `cargo install --list` will show all of them, and then you could parse to check, but not really needed here since `cargo about` will fail with a reasonable message.

eng/psakefile.ps1
```diff
@@ -427,3 +430,42 @@ function Create-DocsEnv() {
...
+    # use cargo-about to generate a notice files
+    # notice files are only for wheel distributions
+    # as no bundled sources are in the sdist.
+
+    # llvm special license is already in the template
+    # as it is a hidden transitive dependency.
+    # https://github.com/EmbarkStudios/cargo-about
+    $config = Join-Path $repo.root notice.toml
+    $template = Join-Path $repo.root notice.hbs
+    Invoke-LoggedCommand -wd $pyqir.parser.dir {
```
cgranade, line 453: Minor, but it feels like these three could be collapsed with a foreach-object over `@($pyqir.parser.dir, $pyqir.generator.dir, $pyqir.evaluator.dir)`?

cgranade (approved):
LGTM!

swernli (approved):

eng/psakefile.ps1
idavis: Done

PR #97, opened 2022-04-05
Adding pyqir-evaluator example with result stream usage.

idavis:


eng/psakefile.ps1
```diff
@@ -397,7 +397,19 @@ task run-examples {
         )
         Assert (@(Compare-Object $bz_first_lines $bz_expected).Length -eq 0) "Expected $bz_expected found $bz_first_lines"
     }
-    
+   
+    exec -workingDirectory $pyqir.evaluator.examples_dir {
+        & $python "teleport.py" | Tee-Object -Variable teleport_output
+        $teleport_first_lines = @($teleport_output | Select-Object -first 5)
```
bettinaheim, line 403: What was the reason to only look at the first case, opposed to comparing all 4 cases?

idavis: They are smoke tests to make sure the samples run without errors and generate some expected output.

examples/evaluator/README.md
```diff
@@ -22,11 +22,30 @@ It consists of a [Python
...
 
   The example can be run using python:
 
   ```bash
   python bernstein_vazirani.py
   ```
+
+- **Teleport example**: <br/>
+This example shows how to log the executed gate sequence leveraging
+a supplied result stream. The `NonadaptiveEvaluator`'s `eval` call accepts
```
bettinaheim, line 36: It may be worth calling out that this functionality is intended for testing purposes. It may otherwise confuse some people that are trying to use this for simulation or connecting to hardware.

eng/psakefile.ps1
bettinaheim: Since the code that we point people to runs for all 4 cases, I think it would make sense to also just include the 4 cases here. It's only a tiny tweak in the build script. I see that the PR is still in draft, so I'll hold off with approving - let me know when it is ready to be reviewed.

idavis: It isn't a simple change. The last bits of the output change based on the result stream. We have an exhaustive teleport tests which verify all of this behavior. The `run-examples` task verifies that the samples run and don't generate errors. If we want to have all of the examples fully verified during the build, we should put that on the backlog as a new task.

bettinaheim: Alright. I realized I never updated my comment after clarifying offline with you whether we have test coverage. As said there, I'll leave the judgement call up to you, as long as we have unit tests for it somewhere.

bettinaheim (approved):

PR #103, opened 2022-04-12
Adding static result allocation.

### LINKED ISSUE #104: Adding static allocation support
## Summary

The code generation should support static and dynamic:
- qubit allocation
- result allocation

The static qubit allocation is the default but is not expressed in the Python APIs. At the moment, only dynamic result allocation is used. This has the side effect that any code which uses dynamic result allocation will not give a target result id during evaluation as the result is returned from the measurement call.

### Generator

- Add `use_static_qubit_alloc(bool)` to SimpleModule
- Add `use_static_result_alloc(bool)` to SimpleModule

The underlying `SemanticModel` defaults to `true` for `use_static_qubit_alloc` and `false` for `use_static_result_alloc`. This is only to avoid a breaking QIR generation change. The initial work will leave the defaults as they are now. When we bump the version number, we will change the default behavior to static result allocation.

### Evaluator

- Add `mz` to the `GateSet` and `GateLogger`
- Open question on whether the `GageLogger` should log `m` or `mz` in the output.

We only know at the semantic model that `mz` was used if we have a result target id.

### Qirlib

Generation:

- Rename `static_alloc` to `use_static_qubit_alloc: bool` on SemanticModel
- Add `use_static_qubit_alloc: bool` on CodeGenerator
- Add `use_static_result_alloc: bool` to SimpleModule
- Add `use_static_result_alloc: bool` to CodeGenerator
- Add support for emitting `mz`: `declare void @__quantum__qis__mz__body(%Qubit*, %Result*)`
- Add generator support for emitting `mz` during static result allocation and `m` during dynamic
- Add support for emitting `read_result`: `declare i1 @__quantum__qis__read_result__body(%Result*)`
- Add generator support for emitting different code for result based branching
  - `read_result` during static result allocation
  - `result_equal` and `get_one` for dynamic as it works right now
- When static result generation is enabled, create int ids for the results 
- Add `requiredResults` metadata with the entrypoint attribute to match the `requiredQubits` emission.

JIT:

- Add `mz` to `GateSet`
- Add extern `mz` decl
- Add static result cache used by extern `mz` decl to record results. This will read from the result stream and insert into the cache based on result id.
- Add extern `read_result` decl: will return the value in the cache for the supplied result id. If the cache doesn't contain that id, it will return `false` (read without measurement)
- Add `reset_static_result_cache` for JIT to clear static cache before module executions.
- Map the extern decls for `mz` and `read_result` to the module if the module declares those extern functions

## Testing

The `pyqir-tests` project will need to run the end-to-end tests with all combinations of qubit/result static/dynamic combination.

More code generation tests will need to be written to verify QIR emission across all combinations of qubit/result static/dynamic.

## Implementation

PR #103

idavis:
Closes #104


qirlib/src/codegen/qis.rs
```diff
@@ -412,4 +457,16 @@ mod qis_declaration_tests {
             str_val.to_string()
         );
     }
+
+    #[test]
+    fn mz_is_declared_correctly() {
```
swernli, line 462: Looks like a matching test for `read_result` is missing.

qirlib/src/generation/emit.rs
```diff
@@ -58,7 +63,7 @@ fn build_entry_function(generator: &CodeGenerator, model: &SemanticModel) -> Res
     add_external_functions(generator, model.external_functions.iter());
     let entry_point = qir::create_entry_point(generator.context, &generator.module);
 
-    if model.static_alloc {
+    if model.use_static_qubit_alloc {
```
swernli, line 66: I can see that this is where the `requiredQubits` attribute is written, but what about the `requiredResults` attribute, is that handled somewhere?

idavis: @swernli didn't know that was needed, I'll add it to the spec issue and get it added

swernli: In case it's helpful, here is the part in QAT that adds it: https://github.com/qir-alliance/qat/blob/685904d5f471bf09b3e5fe5c264adf79ad4bcf09/qir/qat/TransformationRulesPass/TransformationRulesPass.cpp#L741-L746. It uses the same syntax as `requiredQubits`.

swernli (approved):

PR #111, opened 2022-04-21
Adding ability to do musl builds.

### LINKED ISSUE #107: Add musllinux wheel distribution
Alpine (musl) based distribution is not currently supported. To achieve this, several components must be updated:

- [x] Create Alpine OCI base image (maturin, compiler toolchain, etc)
- [x] Get LLVM building correctly in this environment
- [x] Get PyQIR building correctly in this environment
- [x] Update build to create musllinux wheels
- [x] Update build to test musllinux wheels

Wheel should likely follow [PEP 656](https://peps.python.org/pep-0656/) for naming via Maturin with musl 1.2 for the compatibility.

idavis:
Closes #107 

The build has been reworked so that the projects are built the same way. The `musl` and `manylinux` builds are recursive where the outer build reruns the build script in the container. This way all build logic is identical.

The default download logic is now flipped so that qirlib will default to build llvm from source unless downloading is opted in.


Changelog.md
```diff
@@ -12,7 +12,7 @@
 
 - Adding static result allocation by @idavis in https://github.com/qir-alliance/pyqir/pull/103
 - Bumping version to `0.4.0a1` by @idavis in https://github.com/qir-alliance/pyqir/pull/105
-- Changing default value of `use_static_result_alloc` to `True` by @idavis in https://github.com/qir-alliance/pyqir/pull/106
+- Changing default value of use_static_result_alloc to True by @idavis in https://github.com/qir-alliance/pyqir/pull/106
```
swernli, line 15: This seems like an odd update to be in this PR... was this needed for something?

idavis: @swernli  it is a formatting change that was inadvertently added, I'll reset it. It is the same content.

swernli (approved):

Issue #35, opened 2022-01-25
Upgrade to LLVM 13


idavis:
- [ ] Add LLVM build automation
- [ ] Migrate tests/bitcode/ir
- [ ] Update docs
- [ ] Update ENV vars for LLVM_SYS across dev container, code, docs, etc

PR #114, opened 2022-05-05
Add conditional feature support for python bindings and more LLVM versions

nilslice:
I'm working on integrating these crates into some of our QIR work, and it would be nice to not include the Python bindings in places where they are not needed. 

Additionally, at least one of our projects allows the use of LLVM version up to 12 or 13, so this PR also includes more features in each crate that uses `llvm-ir` or `inkwell` to support those versions too. 

All `default-features` have been kept unchanged, so the crates should function precisely the same as if none of the new features were added. 

I've confirmed, at least locally, that our downstream projects work as expected when using/disabling these features.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -65,6 +65,15 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "baf1de4339761588bc0619e3cbc0120ee582ebb74b53b4efbf79117bd2da40fd"
 
 [[package]]
+name = "cmake"
+version = "0.1.48"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "e8ad8cef104ac57b68b89df3208164d228503abbdce70f6880ffa3d970e7443a"
+dependencies = [
+ "cc",
+]
+
+[[package]]
 name = "cty"
 version = "0.2.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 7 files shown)


idavis:
Thanks for the PR @nilslice . This PR would close #35. In order to fully add this, a few more changes are required:

1. `Use-LlvmInstallation` in `utils.ps1` needs to use the correct env variable based on the llvm version
2. Update docs for env vars that need to get set for llvm linking (LLVM_SYS_*)
3. Update `build.rs` with llvm tags based on features that work. Something like:
```rust
fn get_llvm_tag() -> String {
    if let Ok(tag) = env::var("QIRLIB_LLVM_TAG") {
        tag
    } else {
        if cfg!(feature = "llvm-11") {
            "llvmorg-11.1.0".to_owned() // 1fdec59bf
        } else if cfg!(feature = "llvm-12") {
            "llvmorg-12.0.1".to_owned() // fed4134
        } else if cfg!(feature = "llvm-13") {
            "llvmorg-13.0.1".to_owned() // 75e33f7
        } else if cfg!(feature = "llvm-14") {
            "llvmorg-14.0.3".to_owned() // 1f91400
        } else {
            panic!(
                "Unsupported LLVM version. The LLVM feature flags or QIRLIB_LLVM_TAG must be set."
            )
        }
    }
}
```
4. Update the pipelines to create a feature matrix across the LLVM versions setting `QIRLIB_LLVM_TAG` for each build. These versions haven't been tested and we don't know if they will compile and run cleanly.

Additionally, we need to figure out:
- What version of LLVM is the default?
- What version of LLVM is used in the published packages?
- How long do we maintain support for various LLVM versions?

/CC @swernli @bettinaheim

idavis:
@nilslice Based on your PR I have changes with llvm 13 building. I'm merging #111 which has a lot of build cleanup that we can use as a better base for these changes. We have to also bring llvm-sys as an optional dependency for the all-targets-init suppression.

PR #115, opened 2022-05-06
Add bitcode/qir conversion functions to generator

idavis:
This PR adds `ir_to_bitcode` and `bitcode_to_ir` functions to the generator package. Also fixes a bug where the rurned bitcode from `SimpleModule::bitcode()` returned a list of bytes instead of a `bytes` object.


qirlib/src/generation/mod.rs
```diff
@@ -1,3 +1,71 @@
...
+
+use inkwell::{context::Context, memory_buffer::MemoryBuffer};
+
+use crate::module;
+
 pub mod emit;
 pub mod interop;
 pub mod qir;
+
+pub fn ir_to_bitcode(value: &str, name: &str) -> Result<Vec<u8>, String> {
```
swernli, line 12: Interestingly enough, it seems the name is sort of optional (the string is stored at the beginning of the uninitialized memory buffer before any content is read in, and while LLVM C++ objects have functions for getting this identifier, the inkwell code does not). I'm tempted to say this API should skip exposing that and just unconditionally pass the empty string. It seems weird to have a required parameter with an identifier that the user can't then read back out or use without going down several abstraction layers.

swernli: by that same token, I imagine we could decide to remove the name argument from `module::load_memory`, but that would be a breaking change to qirlib at this point.

swernli: After discussing this further, it makes sense that we need this to ensure metadata matches for testing. Thanks for clarifying!

swernli (approved):

PR #119, opened 2022-05-16
Allowing multiple LLVM versions to be used.

idavis:
This PR set up the ability to compile against multiple versions of LLVM. Setting `PYQIR_LLVM_FEATURE_VERSION` to one of the following values `llvm11-0`, `llvm12-0`, and `llvm13-0` will change the LLVM version for all components. `LLVM-IR` doesn't support LLVM 14, so 14 was not added to the build. The default value if the environment variable isn't set is `llvm11-0`.

I have not added and build configurations for CI. The default LLVM version for `qirlib` is not specified, but the build will fail if the features isn't set.

This PR is based off of the work @nilslice did for #114.


idavis:
This does not fully address #35, but makes the change very simple.

idavis:
@nilslice Can you please pull and try again?

nilslice:
@idavis That was it indeed - fixed.

.github/workflows/ci.yml
```diff
@@ -104,7 +104,9 @@ jobs:
         uses: actions/upload-artifact@v2
         with:
           name: manylinux-artifacts
-          path: target/wheels/*manylinux*.whl
+          path: |
+            target/wheels/*manylinux_2_17*.whl
```
swernli, line 108: Just to double check, is this version number informed by a specific version in another part of the repo, like a script or Dockerfile? I didn't see it anywhere else on quick perusal, so wanted to make sure it wasn't labeling something as a specific version when that version might actually change.

eng/utils.ps1
```diff
@@ -260,20 +287,22 @@ function install-llvm {
         [string]$qirlibDir,
         [Parameter(Mandatory)]
         [ValidateSet("download", "build")]
-        [string]$operation
+        [string]$operation,
+        [Parameter(Mandatory)]
+        [ValidateSet("llvm11-0", "llvm12-0", "llvm13-0")]
```
swernli, line 292: The readme mentions that qirlib can compile with LLVM 14, so is it worth adding that to the set or should a version 14 user expect to point to a manually installed/compiled version?

swernli (approved):
Just a couple of questions, but overall looks ready to go!

.github/workflows/ci.yml
idavis: @swernli This version is held by glibc inside of the manylinux container. As we also build on Ubuntu 20.04, it will generate wheels matching `*manylinux_2_31*.whl` which we don't want to distribute. Easier to allowlist than blocklist as versions update.

PR #120, opened 2022-05-17
Support for parsing global byte array constants

swernli:
This change adds support to the pyqir-parser for parsing global byte array constants and getting their defined value from the module. This is useful when dealing global static strings, and the added test demonstrates how to interpret the Python `bytes` as a string.


pyqir-parser/pyqir/parser/_parser.py
```diff
@@ -1,6 +1,7 @@
 # Copyright (c) Microsoft Corporation.
 # Licensed under the MIT License.
 
+from tokenize import String
```
idavis, line 4: @swernli Is this left over from the previous iteration?

pyqir-parser/tests/test_api.py
```diff
@@ -68,6 +68,19 @@ def test_parser_select_support():
     assert isinstance(instr2.false_value, QirLocalOperand)
     assert instr2.false_value.name == "val.i.1"
 
+def test_global_string():
```
idavis, line 71: @swernli  This sample shows reading ascii data converted into utf-8. Does it make sense to have a unicode byte array sample/test?

idavis, line 71: @swernli  This sample shows reading ascii data converted into utf-8. Does it make sense to have a unicode byte array sample/test?

pyqir-parser/pyqir/parser/_parser.py
swernli: It might be... I'll remove it and commit if that doesn't break anything.

pyqir-parser/tests/test_api.py
swernli: Right now we don't have any tools that produce unicode byte arrays, so we'd have to try and author one manually. I'm not sure how to do that at the moment, so this is the only test case we have.

swernli: Looks like this comment somehow showed up twice :).

idavis (approved):

PR #127, opened 2022-06-04
Added mypy github action to check type annotations and mypy stub files [unitaryhack]

### LINKED ISSUE #57: Add mypy to CI
We have type annotations and mypy stub files (.pyi) that aren't checked during CI, so they can drift out of sync with the implementation. We should run mypy during CI to catch these issues. This should be done as a build task run as part of the checks tasks.

WingCode:
closes #57


WingCode:
@idavis I have made the changes you have required.

eng/psakefile.ps1
```diff
@@ -114,6 +117,24 @@ task cargo-clippy -depends init {
...
+    # - Run mypy from within new venv. Reuse same script from task docs
+    Write-Host (Get-ChildItem $wheelhouse -Include *.whl)
+    $envPath = Join-Path $repo.root ".mypy-venv"
+    Create-PyEnv `
+        -EnvironmentPath $envPath `
+        -RequirementsPath (Join-Path $repo.root "eng" "lint-requirements.txt") `
+        -ArtifactPaths (Get-Item $wheelhouse)
+    & (Join-Path $envPath "bin" "Activate.ps1")
+    try {
+        mypy "$($pyqir.parser.python_dir)" "$($pyqir.generator.python_dir)" "$($pyqir.evaluator.python_dir)"
```
idavis, line 131: In order for the build to fail, we have to check the exit code of `mypy`. `Invoke-LoggedCommand` builds on top of `exec` which will do this for us.
```suggestion
        Invoke-LoggedCommand -errorMessage "Please fix the above mypy errors" {
            mypy "$($pyqir.parser.python_dir)" "$($pyqir.generator.python_dir)" "$($pyqir.evaluator.python_dir)"
        }
```

idavis (changes requested):

.github/workflows/link_and_spell_check.yml
```diff
@@ -50,3 +50,8 @@ jobs:
         strict: false
         incremental_files_only: true
         config: 'cspell.json'
+
+  check-type-annotations:
+  - name: "Check type annotations and mypy stub files"
+    run: ./build.ps1 -t checkmypy
+    shell: pwsh
```
idavis, line 57: The check isn't running, I think we need to move this to its own job with steps
```suggestion
  check-type-annotations:
    runs-on: ubuntu-latest
    name: "Check type annotations and mypy stub files"
    steps:
      - uses: actions/checkout@v3
      - name: "Run mypy"
        run: ./build.ps1 -t checkmypy
        shell: pwsh

```

idavis (changes requested):

idavis:
@WingCode mypy requiring the wheels to be built requires extra dependencies. We can try to set up the dependencies in the workflow, but I think it might be better to run it [like the docs](https://github.com/qir-alliance/pyqir/blob/main/.github/workflows/ci.yml#L117-L133). The docs copy the wheels from the arifacts and then run tools which run the wheels.

If we add a task after docs that uses the lines above, and changes the task target to run mypy, I think this will be much easier. What do you think?

WingCode:
Yes @idavis we can do that. We will have to rename [docs](https://github.com/qir-alliance/pyqir/blob/168135555a0cf4fc5b6e786b8c063497e70f3af2/.github/workflows/ci.yml#L117-L133) YAML key to denote it is docs+mypy lint

idavis:
@WingCode I meant copy those lines, paste, and rename the copied job and edit the build target.

WingCode:
@idavis The mypy linting is working as expected as per the logs of failing CI build. Thank you for your patience and support!
Should we fix the mypy lint errors as part of this PR or a different PR?

idavis:
@WingCode The mypy and docs were skipped because the first checks failed. I think we need to remove mypy from the checks list (`task checks -depends cargo-fmt, cargo-clippy, checkmypy`) - Because it depends on rebuild, running the first checks triggers a full build, but on the github actions steps we populate the wheelhouse before running the mypy check which bypasses the full rebuild. I think that once checkmypy is removed from the checks task that everything will work.

I agree that fixing mypy results are a separate task. I'll create an issue for it once the checks run so that we have actionable items.

eng/psakefile.ps1
```diff
@@ -57,7 +60,7 @@ task manylinux -depends build-manylinux-container-image, run-manylinux-container
 
 task musllinux -depends build-musllinux-container-image, run-musllinux-container-image, run-examples-in-musl-containers
 
-task checks -depends cargo-fmt, cargo-clippy
+task checks -depends cargo-fmt, cargo-clippy, checkmypy
```
idavis, line 63:
```suggestion
task checks -depends cargo-fmt, cargo-clippy
```

idavis (changes requested):

idavis (approved):

idavis:
Thanks for the help @WingCode !

PR #128, opened 2022-06-05
[unitaryhack] Evaluator: Raise if bitcode contains unknown external functions.

### LINKED ISSUE #122: pyqir-evaluator should gracefully fail when bitcode contains unknown external functions
**Describe the bug**

When an LLVM file contains declarations of external functions that are unknown, the evaluator crashes without displaying an error message or raising an Exception.

Issue related to #121.

**To Reproduce**

Use an LLVM file that contains unknown external functions, such as:

```
%String = type opaque

declare %String* @__quantum__rt__bool_to_string(i1)

define void @main() #1 {
entry:
  call %String* @__quantum__rt__bool_to_string(i1 1)
  ret void
}

attributes #1 = { "EntryPoint" }
```

This evaluation does not reach the end and throws no Exception:
```
from pyqir.evaluator import GateLogger, NonadaptiveEvaluator

print("Start")

try:
    NonadaptiveEvaluator().eval("CrashMe.ll", GateLogger())
except Exception as e:
    print("An exception was raised.")

print("End")
```

**Expected behavior**

The evaluator should either:
* Handle them in a transparent manner (if possible)
* Display an error message ("Unsupported function...")
* Raise an exception

georgios-ts:
### Summary
This commit updates the evaluator to detect and raise an appropriate error message
if an input LLVM file contains declarations of unknown external functions.

Closes #122


qirlib/src/evaluation/jit.rs
```diff
@@ -116,7 +118,7 @@ fn is_entry_point(function: FunctionValue) -> bool {
         .is_some()
 }
 
-fn module_functions<'ctx>(module: &Module<'ctx>) -> impl Iterator<Item = FunctionValue<'ctx>> {
+pub fn module_functions<'ctx>(module: &Module<'ctx>) -> impl Iterator<Item = FunctionValue<'ctx>> {
```
idavis, line 121: Should probably leave this as `pub(crate)` for now. I'm not sure we want to make this externally visible yet.

qirlib/src/evaluation/jit.rs
```diff
@@ -254,6 +257,23 @@ mod tests {
         Ok(())
     }
 
+    #[serial]
+    #[test]
+    fn fails_if_unknown_external_func() -> Result<(), String> {
```
idavis, line 262: If would be good to see a test in Python to verify that the error comes through correctly.

qirlib/src/evaluation/runtime.rs
```diff
@@ -11,20 +11,28 @@ use log;
...
 }
 
 impl<'ctx> Simulator {
-    pub fn new(module: &Module<'ctx>, ee: &ExecutionEngine<'ctx>) -> Self {
+    pub fn new(module: &Module<'ctx>, ee: &ExecutionEngine<'ctx>) -> Result<Self, String> {
         let simulator = Simulator {
             _scope: crate::evaluation::gates::GateScope::new(),
         };
 
+        if let Some(f) = Simulator::unsupported_function(module) {
```
idavis, line 26: Can we return a list of unsupported functions as string rather than the just the first? From a user perspective it might be helpful to tell them all at once rather than making them re-run each time to find functions that need to be trimmed out.

qirlib/src/evaluation/runtime.rs
```diff
@@ -33,6 +41,15 @@ impl<'ctx> Simulator {
         gs.get_model()
     }
 
+    fn unsupported_function(module: &Module<'ctx>) -> Option<FunctionValue<'ctx>> {
+        module_functions(module).find(|f| {
+            let name = f.get_name().to_str().unwrap();
```
idavis, line 46: Is there a safe way to avoid panicking during the `unwrap` here?

idavis (changes requested):
Overall, this looks good, just a few items I'd like to discuss. Thank you for your hard work.

qirlib/src/evaluation/jit.rs
georgios-ts: Good point, done in 76361cc.

qirlib/src/evaluation/runtime.rs
georgios-ts: Done in 76361cc.

qirlib/src/evaluation/jit.rs
georgios-ts: Done in f65e76f.

qirlib/src/evaluation/runtime.rs
georgios-ts: Since this only fails if a function name contains non-utf8 chars (according to the docs https://doc.rust-lang.org/stable/std/ffi/struct.CStr.html#method.to_str), I assumed that we'd fail before reaching this point (probably while loading the module) and we're safe calling `unwrap` here.

If you're still thinking that this is a problem, we could catch and propagate the error.

pyqir-evaluator/tests/test_unknown_external_func.py
```diff
@@ -0,0 +1,33 @@
...
+            call %String* @__quantum__rt__bool_to_string(i1 1)
+            ret void
+        }
+
+        attributes #1 = { "EntryPoint" }
+    """
+
+    evaluator = NonadaptiveEvaluator()
+    logger = GateLogger()
+    with pytest.raises(OSError):
```
idavis, line 29: Is OSError the correct type for the error?

idavis, line 29: Is OSError the correct type for the error?

pyqir-evaluator/tests/test_unknown_external_func.py
```diff
@@ -0,0 +1,33 @@
+# Copyright (c) Microsoft Corporation.
+# Licensed under the MIT License.
+
+from pyqir.evaluator import GateLogger, NonadaptiveEvaluator
+import pytest
+import tempfile
+
+
```
idavis, line 8: This is looking great. I think we need a test case for use of multiple non-supported calls.

idavis (changes requested):

pyqir-evaluator/tests/test_unknown_external_func.py
georgios-ts: Done in d9741dd

pyqir-evaluator/tests/test_unknown_external_func.py
georgios-ts: It's what we're currently raising. https://github.com/qir-alliance/pyqir/blob/main/pyqir-evaluator/src/python.rs#L101
It's probably not the best choice for every possible error path (e.g multiple entry points in module) but changing this would require keeping information from where the error originated (and not just a string) so we can raise the appropriate error type in Python. I feel it's a bit outside of the scope of this PR.

pyqir-evaluator/tests/test_unknown_external_func.py
```diff
@@ -0,0 +1,64 @@
...
+    logger = GateLogger()
+    with pytest.raises(OSError) as excinfo:
+        with tempfile.NamedTemporaryFile("wt", suffix=".ll") as fd:
+            fd.write(content)
+            fd.flush()
+            evaluator.eval(fd.name, logger)
+    assert str(excinfo.value) == "Unsupported functions `__quantum__rt__bool_to_string`."
+
+
+def test_unknown_external_func():
```
idavis, line 37: This is a redefinition of the above function and needs to be renamed.

idavis (changes requested):

idavis:
@georgios-ts This just needs to be updated as the antlr pr was merged. Then it should be good to go.

idavis (approved):

PR #129, opened 2022-06-06
[unitaryhack] Regenarate mock parser using ANTLR-4.10

### LINKED ISSUE #102: ANTLR 4.10-generated parsers incompatible with previous runtimes
ANTLR released a new version, 4.10, on 11 April 2022. This release breaks the augmented transition network (ATN) serialized state machine in the generated code from earlier versions. The [4.10 Release notes](https://github.com/antlr/antlr4/releases/tag/4.10) explain that they broke the serializer format and the generated code needs to be regenerated with the 4.10 tooling and the python dep needs to be set to 4.10+

georgios-ts:
Closes #102.


examples/generator/mock_language/MockLanguageLexer.py
```diff
@@ -1,36 +1,35 @@
-# Generated from .\MockLanguage.g4 by ANTLR 4.9.1
+# Generated from MockLanguage.g4 by ANTLR 4.10.1
 from antlr4 import *
 from io import StringIO
-from typing.io import TextIO
 import sys
-
+if sys.version_info[1] > 5:
```
idavis, line 5: Can you please add a comment on what/why this is doing? Other than that, everything looks good. Thank you!

idavis (changes requested):

examples/generator/mock_language/MockLanguageLexer.py
georgios-ts: These files are automatically generated from antlr. This check was already present in 4.9.1 version in Parser and introduced since 4.9.2 in Lexer as well https://github.com/antlr/antlr4/pull/3067. The reason was that apparently `typing.io` is deprecated since Python 3.6.

idavis: Sorry, forgot that that whole file is automatically generated.

nathanshammah:
@georgios-ts it seems like you have not signed up on unitaryHACK, please sign up to be able to claim the bounty https://airtable.com/shrJeycewBFqdot2B

PR #133, opened 2022-06-12
Fix type hinting errors

### LINKED ISSUE #132: Fix issues found by mypy checks
PR #127 for issue #57 added checkmypy build task and CI job. That job found issues that need to be resolved. 

```
/home/runner/work/pyqir/pyqir/pyqir-parser/pyqir/parser/_parser.py:4: error: Skipping analyzing "pyqir.parser._native": module is installed, but missing library stubs or py.typed marker
/home/runner/work/pyqir/pyqir/pyqir-parser/pyqir/parser/_parser.py:4: note: See https://mypy.readthedocs.io/en/stable/running_mypy.html#missing-imports
/home/runner/work/pyqir/pyqir/pyqir-parser/pyqir/parser/_parser.py:607: error: Incompatible types in assignment (expression has type "QirType", variable has type "None")
/home/runner/work/pyqir/pyqir/pyqir-parser/pyqir/parser/_parser.py:608: error: Incompatible return value type (got "None", expected "QirType")
/home/runner/work/pyqir/pyqir/pyqir-parser/pyqir/parser/_parser.py:940: error: Incompatible types in assignment (expression has type "List[QirInstr]", variable has type "None")
/home/runner/work/pyqir/pyqir/pyqir-parser/pyqir/parser/_parser.py:941: error: Incompatible return value type (got "None", expected "List[QirInstr]")
/home/runner/work/pyqir/pyqir/pyqir-parser/pyqir/parser/_parser.py:950: error: Incompatible types in assignment (expression has type "QirTerminator", variable has type "None")
/home/runner/work/pyqir/pyqir/pyqir-parser/pyqir/parser/_parser.py:951: error: Incompatible return value type (got "None", expected "QirTerminator")
/home/runner/work/pyqir/pyqir/pyqir-parser/pyqir/parser/_parser.py:962: error: Incompatible types in assignment (expression has type "List[QirPhiInstr]", variable has type "None")
/home/runner/work/pyqir/pyqir/pyqir-parser/pyqir/parser/_parser.py:963: error: Incompatible return value type (got "None", expected "List[QirPhiInstr]")
/home/runner/work/pyqir/pyqir/pyqir-parser/pyqir/parser/_parser.py:998: error: Incompatible types in assignment (expression has type "QirType", variable has type "None")
/home/runner/work/pyqir/pyqir/pyqir-parser/pyqir/parser/_parser.py:999: error: Incompatible return value type (got "None", expected "QirType")
/home/runner/work/pyqir/pyqir/pyqir-parser/pyqir/parser/_parser.py:1028: error: Incompatible types in assignment (expression has type "List[QirParameter]", variable has type "None")
/home/runner/work/pyqir/pyqir/pyqir-parser/pyqir/parser/_parser.py:1029: error: Incompatible return value type (got "None", expected "List[QirParameter]")
/home/runner/work/pyqir/pyqir/pyqir-parser/pyqir/parser/_parser.py:1037: error: Incompatible types in assignment (expression has type "QirType", variable has type "None")
/home/runner/work/pyqir/pyqir/pyqir-parser/pyqir/parser/_parser.py:1038: error: Incompatible return value type (got "None", expected "QirType")
/home/runner/work/pyqir/pyqir/pyqir-parser/pyqir/parser/_parser.py:1046: error: Incompatible types in assignment (expression has type "List[QirBlock]", variable has type "None")
/home/runner/work/pyqir/pyqir/pyqir-parser/pyqir/parser/_parser.py:1047: error: Incompatible return value type (got "None", expected "List[QirBlock]")
/home/runner/work/pyqir/pyqir/pyqir-evaluator/pyqir/evaluator/_nonadaptiveevaluator.py:5: error: Skipping analyzing "pyqir.evaluator._native": module is installed, but missing library stubs or py.typed marker
Found 18 errors in 2 files (checked 10 source files)
```

LaurentAjdnik:
Fixes #132 by adding type hints.

### DIFF
--- a/pyqir-parser/pyqir/parser/_parser.py
+++ b/pyqir-parser/pyqir/parser/_parser.py
@@ -603,13 +603,9 @@ class QirInstr:
         Gets the QirType instance representing the output of this instruction. If the instruction
         has no output, the type will be an instance of QirVoidType.
         """
-        if self._type is None:
-            ty = QirType(self.instr.type)
-            self._type = ty
-        else:
-            ty = self._type
-        return ty
-
+        if self._instructions is None:
+            self._instructions = [QirInstr(i) for i in self.block.instructions]
+        return cast(List[QirInstr], self._instructions)
 
 class QirOpInstr(QirInstr):
     """
@@ -949,7 +945,7 @@ class QirBlock:
         Gets the terminator instruction for this block. Every block has exactly one terminator
         and it is the last intruction in the block.
         """
-        if self._terminator == None:
+        if self._terminator is None:
             self._terminator = QirTerminator(self.block.terminator)
         return cast(QirTerminator, self._terminator)
 
@@ -961,7 +957,7 @@ class QirBlock:
         of phi nodes, but they are always the first instructions in any given block. A block with no
         phi nodes will return an empty list.
         """
-        if self._phi_nodes == None:
+        if self._phi_nodes is None:
             self._phi_nodes = [QirPhiInstr(i) for i in self.block.phi_nodes]
         return cast(List[QirPhiInstr], self._phi_nodes)
 
@@ -997,7 +993,7 @@ class QirParameter:
         """
         Gets the type of this parameter as represented in the QIR.
         """
-        if self._type == None:
+        if self._type is None:
             self._type = QirType(self.param.type)
         return cast(QirType, self._type)
 
@@ -1026,7 +1022,7 @@ class QirFunction:
         """
         Gets the list of parameters used when calling this function.
         """
-        if self._parameters == None:
+        if self._parameters is None:
             self._parameters = [QirParameter(i) for i in self.func.parameters]
         return cast(List[QirParameter], self._parameters)
 
@@ -1035,7 +1031,7 @@ class QirFunction:
         """
         Gets the return type for this function.
         """
-        if self._return_type == None:
+        if self._return_type is None:
             self._return_type = QirType(self.func.return_type)
         return cast(QirType, self._return_type)
 
@@ -1044,7 +1040,7 @@ class QirFunction:
         """
         Gets all the basic blocks for this function.
         """
-        if self._blocks == None:
+        if self._blocks is None:
             self._blocks = [QirBlock(i) for i in self.func.blocks]
         return cast(List[QirBlock], self._blocks)
 
@@ -1120,7 +1116,7 @@ class QirModule:
         """
         Gets all the functions defined in this module.
         """
-        if self._functions == None:
+        if self._functions is None:
             self._functions = [QirFunction(i) for i in self.module.functions]
         return self._functions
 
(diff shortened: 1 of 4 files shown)


pyqir-parser/pyqir/parser/_parser.py
```diff
@@ -605,7 +605,7 @@ def type(self) -> QirType:
         """
         if self._type == None:
             self._type = QirType(self.instr.type)
-        return self._type
+        return cast(QirType, self._type)
```
samarsha, line 608: I wonder if we can use [https://docs.python.org/3/library/functools.html#functools.cached_property](`@functools.cached_property`) instead? That would remove the extra cache fields and the casts.

samarsha, line 608: I wonder if we can use [https://docs.python.org/3/library/functools.html#functools.cached_property](`@functools.cached_property`) instead? That would remove the extra cache fields and the casts.

samarsha: Hmm, never mind, `cached_property` is from Python 3.8... It would still be nice to avoid `cast` in too many places since it can be used unsoundly, but that's probably out of scope for this PR.

samarsha: Another option is to refactor this to something like:

```python
if self._type is None:
    ty = QirType(self.instr.type)
    self._type = ty
else:
    ty = self._type
return ty
```

where Mypy should hopefully be able to convince itself that `ty` is not an Optional.

LaurentAjdnik: Thanks @samarsha for your feedback!

Just tested the refactoring. It went through.

But I'm not a big fan of this workaround:

- The code is longer
- At first sight, it does not look efficient/straightforward
- It introduces a new unnecessary variable
- This variable itself is not type-hinted

Why do you think that cast "can be used unsoundly" ?

It is only used for type hinting.

On a personal standpoint, I'd favor the "unrefactored" version.

But I'll follow the community's conclusions, of course. 😅

samarsha: But I recognize also that the code was already using this pattern before the type hints were added, so if you don't want to refactor in this PR, I'd be OK with using `cast` for now.

LaurentAjdnik: Sorry for not replying earlier.

Very instructive demonstration, thanks!

You make a point and it's a pity it adds complexity. Well, `mypy` often does...

Not having much time to correct all occurrences, I will get back to the previous version with `cast`, since it's OK for you.

It will allow us to close #132 and the specific update of removing `cast`s could be another issue/PR.

LaurentAjdnik: PS : I also replaced all `== None` (including elsewhere in `_parser.py`) with `is None`.

samarsha (approved):
Thanks!

PR #137, opened 2022-07-13
Parser should throw appropriate exception when loading invalid bitcode

### LINKED ISSUE #136: Parser should throw appropriate exception when loading invalid bitcode
The parser currently crashes python when loading non-bitcode files. The API should validate that a bitcode file is being loaded and throw an appropriate Python error.

idavis:
Fixes #136 

llvm-ir error handling doesn't work correctly and crashes python. This change uses Inkwell to load the bitcode and bubble up errors. If Inkwell can load the module, we let llvm-ir load the bitcode as normal.

I tried to fix this using just llvm-ir, but they made the classes and methods needed to implement the fix private. So short of a PR, there was no way to create a fix.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -477,6 +477,7 @@ dependencies = [
 name = "pyqir-parser"
 version = "0.5.0-alpha"
 dependencies = [
+ "inkwell",
  "llvm-ir",
  "pyo3",
 ]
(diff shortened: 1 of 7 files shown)


samarsha (approved):

swernli (approved):
Good catch and fix! If you haven't already, it would be worth filing an issue against llvm_ir so they are aware of the problem.

PR #138, opened 2022-07-20
Support functions with return values

samarsha:
This PR adds support for calling external functions that have a return value, and using that return value as the argument to other external functions or rotation gates. It adds a variable case to PyQIR values. Variables are untyped (type errors will be caught only by LLVM).


pyqir-generator/pyqir/generator/_values.py
```diff
@@ -1,11 +1,11 @@
...
-A QIR or LLVM value, or a Python value that can be automatically converted into
-one.
-"""
+# Copyright (c) Microsoft Corporation.
+# Licensed under the MIT License.
+
+from pyqir.generator._native import Qubit, ResultRef, Value
+from typing import Union
+
+Valuable = Union[Value, bool, int, float, Qubit, ResultRef]
```
samarsha, line 7: The Value type is the true representation of a value, and it's a Python wrapper around a Rust enum for each type of value (including variables). The rest of the Python types we support automatically converting into a Value for convenience. "Valuable" is just my cute name for "able to be (converted into) a Value", but let me know if you have a better name.

pyqir-generator/tests/test_external_functions.py
```diff
@@ -203,3 +203,74 @@ def test_wrong_number_of_args(self) -> None:
...
+    def test_variable_wrong_external_type(self) -> None:
+        mod = SimpleModule("test", 0, 0)
+        foo = mod.add_external_function("foo", types.Function([], types.INT))
+        bar = mod.add_external_function(
+            "bar", types.Function([types.QUBIT], types.VOID))
+
+        x = mod.builder.call(foo, [])
+        mod.builder.call(bar, [x])
+
+        with self.assertRaisesRegex(OSError, "Call parameter type does not match function signature!"):
```
samarsha, line 229: To be honest I'm not sure where the OSError is coming from specifically, if we wanted to change it to another exception type. It seems like it's consistent with other LLVM errors that raise OSError, though.

samarsha:
There might be some churn on some of the things added in this PR with #140 (for example, I needed to make variables typed there), but that could change depending on how the review for #140 goes.

samarsha:
I removed Valuable and added explicit constructors for constant values. ResultRef is kind of an odd case. I want to remove it and absorb it into Value, but it's not a proper value when static result allocation is turned off because of the magic mapping. So I'll defer that to another PR.

pyqir-generator/pyqir/generator/_native.pyi
```diff
@@ -29,30 +29,55 @@ def bitcode_to_ir(bitcode: bytes, module_name: Optional[str], source_file_name:
...
     ...
 
 
 class Function:
-    """A callable value for a module function."""
+    """A QIR function."""
     ...
 
 
+class Value:
```
Roland-djee, line 42: A point that is slightly confounding to me is the multiple definition of the `Value` type. For instance, here, it looks like a module-wide definition for the purpose of holding either an integer or a double type for (not only) qubit indexing and gate parametrisation typing. This is to be contrasted with the definition in `types.py` (see [here](https://github.com/qir-alliance/pyqir/blob/38dd3d0540a646588b1ed3cfc053de3097581dc8/pyqir-generator/pyqir/generator/types.py#L42)) which looks like the type exposed to the user API.

pyqir-generator/tests/test_external_functions.py
```diff
@@ -128,11 +137,11 @@ def test_call_two_dynamic_results(self) -> None:
         )
 
     def test_call_numbers(self) -> None:
-        b = True
+        b = Value.integer(types.BOOL, True)
```
Roland-djee, line 140: Is it necessary to do this ? It looks like an overcomplicated and confusing syntax to me. Would defining a `bool` field in `Value` make it easier to use ?

Roland-djee (commented):
Thanks for this work really. Looks good to me. Only two minor comments for your consideration. I haven't gone in detail through the Rust code but I would assume my comments might apply there too.

pyqir-generator/pyqir/generator/_native.pyi
samarsha: - `pyqir.generator._native.Value` is a QIR/LLVM value.
- `pyqir.generator.types.Value` is a QIR/LLVM type, restricted to the set of types that a value can be (that is, excluding void).

I agree that the naming is confusing, though. I think we don't actually need to keep the void type separate, so I combined `types.Value` and `types.Return` into one thing called `Type`.

pyqir-generator/tests/test_external_functions.py
samarsha: We could define `Value.bool` as a shorthand for creating an `i1` value.

In general, PyQIR needs a way to create values of any integer type. For example `Value.integer(types.Integer(16), 1)` for a 16-bit integer. Would you prefer to have more concise syntax for all integer values in addition to bools?

Roland-djee: Well, that is a matter of taste but in general I am in favour of conciseness. I would assume that passing the width directly to `Value.integer` for any non default 32-bit integer would make it easier to use.

pyqir-generator/pyqir/generator/_native.pyi
Roland-djee: Thanks for the explanation. At the risk of being a bit pedantic, it looks to me that this is redundant now with the use of types in the `types` module with the exception of `BOOL`. Just a suggestion: would it make sense to keep the distinction and name them `ValueType` and `ReturnType` ?

samarsha: Sorry, I'm not sure what you mean by `Value` being redundant, could you explain?

pyqir-generator/tests/test_external_functions.py
samarsha: Something to consider is that we're planning in the near future to transition PyQIR from being a deep embedding of LLVM to a shallow one. At that point, we won't be able to create constant values without a reference to the LLVM context (which all types have). We're trying to design the current API so that this switch over will cause as few breaking changes as possible. So I'd prefer to keep requiring a type object for constant value creation.

I took another stab at making the syntax more concise given this constraint, though. The new syntax is:

```
const(types.BOOL, True)
const(types.Int(16), 123)
```

Roland-djee: Looks good to me !

pyqir-generator/pyqir/generator/_native.pyi
Roland-djee: Ah, sorry I've been unclear. I meant the fact that QIR/LLVM types are defined [here](https://github.com/qir-alliance/pyqir/blob/ed8c20b2a216a7e0e79a49964b24e7692bf4e191/pyqir-generator/pyqir/generator/types.py#L53) but also accessed through the aliases (`types.<type>`). In short, I am wondering if the definition of aliases is necessary. But there might be a need that eludes me now.

samarsha: Do you mean these aliases?

```python
VOID: Void = Void._VOID
DOUBLE: Double = Double._DOUBLE
QUBIT: Qubit = Qubit._QUBIT
RESULT: Result = Result._RESULT
```

That's an artifact of how I'm using `Union` to define a sum type - the nullary constructors are awkward. Making them singleton types seemed like the least worst option, but I didn't want to make people write `Double.DOUBLE`, so I defined module-level aliases for the singleton instances.

idavis (approved):

pyqir-generator/pyqir/generator/_native.pyi
Roland-djee: Ah OK. I understand now. This is because when `types.Function` is instantiated, it takes as argument a list of references to types instances. Thanks.

samarsha: The issue exists even if the type is not recursive. For example, how can this Rust type be encoded in Python:

```rust
enum Sum {
    Foo,
    Bar { value: i64 },
}
```

As a union, each case is its own type and could be encoded like:

```python
@dataclass
class Foo: pass

@dataclass
class Bar:
    value: int

Sum = Foo | Bar
```

But now you have to construct and pattern match on `Foo()`, not `Foo` (the parentheses are annoying), and each time you construct `Foo()` a new object in memory is instantiated (which is unnecessary). So I make `Foo` a singleton using `Enum`:

```python
class Foo(Enum):
    FOO = auto()
```

but still `Foo.FOO` is the member of `Sum`, not `Foo` itself. So I add the module-level alias to `Foo`'s instance for convenience.

Roland-djee: Yep, I get that. I was just wondering if doing something as simple as:

```python
class Sum(Enum):
     Foo
     Bar: int
```
Wouldn't allow to do what you want.

samarsha: In Python, enum cases are like static fields, not constructors. So every case needs to be of the form `name = value`. `Sum.Bar` isn't expressible as an enum case because you can't construct `Sum.Bar(x)` for an arbitrary *x*. In the definition you gave, `Foo` and `Bar` are uninitialized attributes and can't be used.

PR #140, opened 2022-07-22
Support LLVM intrinsic operators for integers

samarsha:
This PR uses the variables feature in #138 to support several LLVM intrinsic operators for integers in the instruction builder.


qirlib/src/generation/interop.rs
```diff
@@ -1,6 +1,8 @@
 // Copyright (c) Microsoft Corporation.
 // Licensed under the MIT License.
 
+pub use inkwell::IntPredicate;
```
samarsha, line 4: Might be better to redefine this instead of exposing the Inkwell type?

pyqir-generator/pyqir/generator/_native.pyi
```diff
@@ -58,13 +59,104 @@ def const(ty: Type, value: Union[int, float]) -> Value:
...
+        """
+        Inserts an addition instruction.
+
+        :param lhs: The left-hand side.
+        :param rhs: The right-hand side.
+        :returns: The sum.
+        """
+        ...
+
+    def sub(self, lhs: Value, rhs: Value) -> Value:
```
samarsha, line 102: `builder.neg(x)` is not supported directly, but is equivalent to `builder.sub(const(your_int_type, 0), x)`.

pyqir-generator/pyqir/generator/_native.pyi
```diff
@@ -58,13 +59,104 @@ def const(ty: Type, value: Union[int, float]) -> Value:
...
+        """
+        Inserts a logical (zero fill) shift right instruction.
+
+        :param lhs: The value to shift.
+        :param rhs: The number of bits to shift by.
+        :returns: The result.
+        """
+        ...
+
+    def icmp(self, pred: IntPredicate, lhs: Value, rhs: Value) -> Value:
```
samarsha, line 142: Depending on what users are planning to do with `icmp`, we might need to add a branching instruction to the builder, like `BasicQisBuilder.if_result` but conditioned on a boolean instead of a result.

qirlib/src/generation/interop.rs
idavis: @samarsha I agree, I we shouldn't reexport inkwell out of the interop layer

qirlib/src/generation/qir/instructions.rs
```diff
@@ -157,11 +157,33 @@ pub(crate) fn emit<'ctx>(
...
+) {
+    let lhs = get_value(generator, env, &op.lhs).into_int_value();
+    let rhs = get_value(generator, env, &op.rhs).into_int_value();
+    let result = match op.kind {
+        BinaryKind::And => generator.builder.build_and(lhs, rhs, ""),
+        BinaryKind::Or => generator.builder.build_or(lhs, rhs, ""),
+        BinaryKind::Xor => generator.builder.build_xor(lhs, rhs, ""),
+        BinaryKind::Add => generator.builder.build_int_add(lhs, rhs, ""),
+        BinaryKind::Sub => generator.builder.build_int_sub(lhs, rhs, ""),
+        BinaryKind::Mul => generator.builder.build_int_mul(lhs, rhs, ""),
```
idavis, line 179: @samarsha Do we have plan for dealing with this when we have float/fixed support? Calling `.into_int_value()` won't work for those ops in the future.

samarsha:
@Roland-djee FYI, this one is ready for review.

qirlib/src/generation/qir/instructions.rs
samarsha: Floating-point operators have different names (fadd, fsub, fmul, etc.). These operators will always be for integers only.

Fixed-points are represented as integers so they can use integer add/sub. Fixed-point mul and div intrinsics also have different names in LLVM.

qirlib/src/generation/qir/instructions.rs
```diff
@@ -157,11 +157,52 @@ pub(crate) fn emit<'ctx>(
...
+        IntPredicate::UGT => inkwell::IntPredicate::UGT,
+        IntPredicate::UGE => inkwell::IntPredicate::UGE,
+        IntPredicate::ULT => inkwell::IntPredicate::ULT,
+        IntPredicate::ULE => inkwell::IntPredicate::ULE,
+        IntPredicate::SGT => inkwell::IntPredicate::SGT,
+        IntPredicate::SGE => inkwell::IntPredicate::SGE,
+        IntPredicate::SLT => inkwell::IntPredicate::SLT,
+        IntPredicate::SLE => inkwell::IntPredicate::SLE,
+    }
+}
```
idavis, line 204: @samarsha Does it make sense to implement `Into` here instead?

pyqir-generator/pyqir/generator/_native.pyi
samarsha: I'll do this in a separate PR.

examples/generator/arithmetic.py
```diff
@@ -0,0 +1,30 @@
...
+# Declare functions that can produce and consume integers at runtime. See
+# external_functions.py.
+i32 = types.Int(32)
+get_int = mod.add_external_function("get_int", types.Function([], i32))
+take_int = mod.add_external_function(
+    "take_int", types.Function([i32], types.VOID))
+
+# Add 3 to a number and multiply the result by 2.
+a = mod.builder.call(get_int, [])
+b = mod.builder.add(const(i32, 3), a)
```
idavis, line 18: @samarsha I think we should pull `const(i32, 3)` here out and explain why `const` is being called.

idavis (approved):

Roland-djee:
> @Roland-djee FYI, this one is ready for review.

Thank you for this @samarsha . I am now back from holidays and I'll be happy to be a bit more interactive. At first glance, it looks good to me. Let me try that out in `pytket-qir` and feedback you on it.

PR #145, opened 2022-08-19
Support branching on boolean conditions

samarsha:
To provide a use case for `icmp` from #140, add `if`, which is like `if_result` but conditioned on a boolean instead of a measurement result.


qirlib/src/generation/qir/instructions.rs
```diff
@@ -234,20 +237,53 @@ fn emit_if<'ctx>(
...
+        generator,
+        env,
+        entry_point,
+        cond,
+        &if_result.then_insts,
+        &if_result.else_insts,
+    );
+}
+
+fn emit_if_inkwell<'ctx>(
```
samarsha, line 273: this needs a better name

idavis (approved):

Issue #143, opened 2022-08-01
Add custom attributes to functions


idavis:
In the Python API, add the ability to add custom attribute metadata when defining functions.

PR #147, opened 2022-08-23
Remove dynamic allocation for qubits and results

### LINKED ISSUE #53: Clean up qirlib generator API to match current Python API
The qirlib generator API contains some older design choices that don't match the new pyqir-generator API (#32), and should be reconsidered:

- Instructions use string identifiers for results and qubits, but the new Python API abstracts away the string. It would be simpler and cheaper to use numeric identifiers.
- `QuantumRegister` represents a single qubit, but `ClassicalRegister` represents a sequence of results, which is asymmetric. The Python API treats them symmetrically.

samarsha:
Closes #53.


qirlib/src/generation/emit.rs
```diff
@@ -67,40 +59,24 @@ fn build_entry_function(generator: &CodeGenerator, model: &SemanticModel) -> Res
...
-    if generator.use_static_result_alloc {
-        let num_results: u64 = model.registers.iter().map(|f| f.size).sum();
-        let num_result_string = format!("{}", num_results);
+    if model.num_results > 0 {
         let required_results = generator
             .context
-            .create_string_attribute("requiredResults", &num_result_string);
+            .create_string_attribute("requiredResults", &model.num_results.to_string());
         entry_point.add_attribute(AttributeLoc::Function, required_results);
     }
```
samarsha, line 74: I'm not sure if:
- the `requiredQubits` and `requiredResults` attributes should be omitted if the count is 0; or
- the attributes should always be present; or
- `num_qubits` and `num_results` in `SemanticModel` should be `Option<u64>`.

qirlib/src/generation/qir/instructions.rs
```diff
@@ -303,3 +234,45 @@ fn emit_if<'ctx>(
...
+    let value = generator.u64_to_i64(id).into_int_value();
+    let ty = generator.qubit_type().ptr_type(AddressSpace::Generic);
+    generator.builder.build_int_to_ptr(value, ty, "")
+}
+
+fn get_result<'ctx>(generator: &CodeGenerator<'ctx>, id: u64) -> PointerValue<'ctx> {
+    let value = generator.u64_to_i64(id).into_int_value();
+    let ty = generator.result_type().ptr_type(AddressSpace::Generic);
+    generator.builder.build_int_to_ptr(value, ty, "")
+}
```
samarsha, line 269: It's allowed to use static qubit and result IDs that are outside the range declared by the `requiredQubits` and `requiredResults` attributes. I'm not sure if we should stop people from doing that.

swernli: Based on our conversation, it is apparent that the attributes are set directly by the user without any checks, and get overwritten by processing in QAT anyway, so we should not worry about emitting these at all. They code related to qubit/result count attribute handling can be removed.

qirlib/src/generation/emit.rs
swernli: Based on the other answer, I think this code can just be removed.

swernli (commented):
I can re-review once the qubit/result count attributes are removed.

qirlib/src/generation/qir/instructions.rs
samarsha: Sounds good, I prefer removing the attributes too. FYI @idavis

idavis: @swernli @samarsha FYI, for the future: https://github.com/qir-alliance/pyqir/issues/143

swernli (approved):

PR #152, opened 2022-08-27
Improve Mypy checking in CI

samarsha:
The CI was not running Mypy on most of our Python files, and it was not using strict mode. This PR fixes both of those issues and also makes it easier to use Mypy locally by adding a configuration file.

I actually found a bug in pyqir-parser using mypy: QirArrayType defined a property `element_types(self) -> List[QirType]` but it should have been `element_type(self) -> QirType`.


pyqir-generator/tests/test_conversions.py
```diff
@@ -13,22 +13,22 @@ def get_module() -> SimpleModule:
...
+    converted_ir = bitcode_to_ir(bitcode, "test", None)
     assert expected_ir == converted_ir
 
 
 def test_ir_round_trip_is_not_identical_when_module_name_isnot_supplied() -> None:
     expected_ir = get_module().ir()
-    bitcode = ir_to_bitcode(expected_ir)
-    converted_ir = bitcode_to_ir(bitcode)
+    bitcode = ir_to_bitcode(expected_ir, None, None)
+    converted_ir = bitcode_to_ir(bitcode, None, None)
```
swernli, line 24: For this change, is the `None` required because of how type-checked this file is or will other python consumers now have to explicitly include `None` arguments? If it's the latter, then this may end up being a breaking change for things like qiskit to QIR test infra.

swernli (approved):
One question about API changes (from context I assume this isn't a breaking change), otherwise looks good!

pyqir-generator/tests/test_conversions.py
samarsha: This file wasn't type checked in CI before. I didn't change the type of `ir_to_bitcode` and `bitcode_to_ir` - they were already type hinted as having three required parameters (although apparently the implementation allows default arguments).

samarsha: I can change the type instead to allow default arguments, that probably looks nicer for callers.

PR #157, opened 2022-09-22
Replace SemanticModel with Inkwell in PyQIR Generator

samarsha:
This is a rewrite of pyqir-generator and qirlib to remove the SemanticModel representation, and use Inkwell/LLVM types directly. This is the first step towards unifying pyqir-parser and pyqir-generator.


pyqir-generator/tests/test_external_functions.py
```diff
@@ -125,61 +168,77 @@ def test_call_numbers(self) -> None:
...
-                f = mod.add_external_function(
-                    "test_function", types.Function([types.BOOL], types.VOID)
-                )
+    def test_overflow_int_value(self) -> None:
+        mod = SimpleModule("test", 0, 0)
+        types = mod.types
+        f = mod.add_external_function("f", types.function(types.void, [types.int(32)]))
+        i = const(mod.types.int(32), 2**32 + 123)
+        mod.builder.call(f, [i])
+        self.assertIn("call void @f(i32 123)", mod.ir())
```
idavis, line 212: Does it make sense to add test with negative values?

qirlib/src/tests.rs
```diff
@@ -0,0 +1,69 @@
...
+use std::{env, fs, path::PathBuf};
+
+/// Compares generated IR against reference files in the "resources/tests" folder. If changes
+/// to code generation break the tests:
+///
+/// 1. Run the tests with the `PYQIR_TEST_SAVE_REFERENCES` environment variable set to
+///    regenerate the reference files.
+/// 2. Review the changes and make sure they look reasonable.
+/// 3. Unset the environment variable and run the tests again to confirm that they pass.
+pub(crate) fn assert_reference_ir(
```
idavis, line 16: Should this use #[cfg(test)] so that is isn't shipped in the binary?

pyqir-generator/Cargo.toml
```diff
@@ -10,6 +10,7 @@ homepage = "https://github.com/qir-alliance/pyqir"
 repository = "https://github.com/qir-alliance/pyqir"
 
 [dependencies]
+inkwell = { git = "https://github.com/TheDan64/inkwell", branch = "master", default-features = false, features = ["target-x86"] }
 pyo3 = { version = "0.17.2", features = ["abi3-py37", "extension-module"] }
 qirlib = { path = "../qirlib" }
```
idavis, line 16: I think there was a bad merge here. We see pyo3 dep from old branch instead of new base. Same with features block below.

qirlib/src/tests.rs
samarsha: This module in its entirety is `#[cfg(test)]` in lib.rs.

pyqir-generator/Cargo.toml
samarsha: Sorry, I'm not sure what you mean. The only change to this file is the addition of the inkwell dependency, which is intentional.

pyqir-generator/tests/test_external_functions.py
samarsha: Sure, added. It turns out that negative values are not supported at all. Probably because Inkwell `const_int` takes a u64 value and the PyO3 conversion to unsigned fails if negative.

idavis (approved):

Issue #135, opened 2022-07-06
Add LLVM 14 support


idavis:
Prerequisites:
 - #35 
 - https://github.com/cdisselkoen/llvm-ir/issues/18

PR #170, opened 2022-10-22
Add Inkwell-based parsing

### LINKED ISSUE #77: Parser should allow construction from in-memory bitcode
Currently, the pyqir_parser only supports initializing the parser from a string that contains the path to an on-disk bitcode file. It would be helpful to support constructing a parser instance from an in-memory bitcode file, as that would open up more scenarios where code is already actively operating on bitcode, such as receiving it as a network stream.

Note that this may require updates to the underlying `llvm_ir` module that the parser depends on.

### LINKED ISSUE #151: Don't use namespace packages
Namespace packages have poor tooling support. We are stuck using old pkgutil-style namespace packages because Maturin doesn't and won't support the new native namespace packages (PyO3/maturin#811). Meanwhile, mypy supports only native namespace packages and not pkgutil-style ones (python/mypy#9393).

I think we should use `pyqir_generator`, `pyqir_parser`, `pyqir_evaluator` instead of `pyqir.generator`, `pyqir.parser`, `pyqir.evaluator`. It's simpler and just works in more scenarios.

### LINKED ISSUE #175: Unify parsing and code generation
PyQIR's parsing functionality and code generation functionality are currently totally independent; parsed QIR is read-only and can't be edited and re-emitted, and generated QIR is write-only. This precludes use cases like doing simple transforms or optimizations with PyQIR on QIR that came from another source. It's also more complicated to maintain since there is overlap between the two, and we're using a different Rust LLVM wrapper for each.

We want to combine Parser and Generator into one cohesive thing that can do both. The first step was refactoring Generator to use LLVM objects directly in #157. Next we will add parsing functionality to the LLVM-backed Generator types in #174. Finally, the pyqir-parser package will be retired. We will transition to a single pyqir package that has the same functionality as pyqir-parser, pyqir-generator, and pyqir-evaluator (with the first step in #170).

For users of pyqir-generator and pyqir-evaluator, the impact will be minimal. The main breaking change is to how types for external functions are created, which came with #157. pyqir-parser users will need to adapt to a completely rewritten API, although it will support the same functionality.

A sketch of the proposed API for the combined parsing and generating functionality is here: https://gist.github.com/samarsha/abdfd9fa7b6aee133904935c7b4511a5

The changes to external functions look like this. Before:

```python
from pyqir.generator import SimpleModule, types

m = SimpleModule("foo", 1, 1)
f = m.add_external_function(
    "f",
    types.Function([types.Int(8), types.QUBIT], types.VOID)
)
```

After:

```python
from pyqir.generator import SimpleModule

m = SimpleModule("foo", 1, 1)
types = m.types
f = m.add_external_function(
    "f",
    types.function(types.void, [types.int(8), types.qubit])
)
```

samarsha:
Closes #77, closes #151, closes #175.


idavis:
Closes #135

idavis (approved):

PR #201, opened 2022-11-21
Add null constants, `br` instruction, and clean up API

samarsha:
- Add `Constant.null(Type)`.
- Add `Builder.br(BasicBlock)`.
- Rename `constant_bytes` to `extract_bytes`.


pyqir/src/values.rs
```diff
@@ -113,36 +114,33 @@ impl Value {
 ///
 /// :param Context context: The global context.
```
idavis, line 115: Please clarify the expected behavior based on the parameters.

pyqir/src/values.rs
```diff
@@ -385,6 +409,40 @@ impl From<Linkage> for inkwell::module::Linkage {
...
+impl ConstantExpr {
+    /// Creates a `getelementptr` (GEP) constant expression.
+    ///
+    /// :param Value value: The aggregate value.
+    /// :param Sequence[Value] indices: The indices of the element.
+    /// :param bool inbounds: Whether to create an in-bounds GEP.
+    /// :returns: The GEP constant expression.
+    /// :rtype: ConstantExpr
+    #[staticmethod]
+    #[allow(clippy::needless_pass_by_value)]
```
idavis, line 425: Please add a test for inbounds == false

pyqir/tests/test_entry_points.py
```diff
@@ -84,3 +84,19 @@ def test_invalid_module() -> None:
         verify_module(module)
         == "Basic Block in function 'main' does not have terminator!\nlabel %0\n"
     )
+
+
+def test_multiple_blocks() -> None:
```
idavis, line 89: Please add tests for the matrix of options

idavis (changes requested):

pyqir/src/values.rs
samarsha: For some reason, LLVM doesn't seem to care if I use inbounds or not - it emits `getelementptr inbounds` no matter what... so I just removed the parameter.

idavis (approved):

PR #204, opened 2022-11-24
Keep the module alive

### LINKED ISSUE #196: Function objects can outlive their parent module
If a function object outlives its parent module, like in this example:

```python
def foo(bitcode: bytes) -> Function:
    return Module.from_bitcode(bitcode).functions[0]
```

then the returned function is invalid, for example `str(foo(bitcode))` can segfault.

samarsha:
Closes #196.


pyqir/src/values.rs
```diff
@@ -75,37 +76,110 @@ impl Value {
...
+            Self::Context(context) => context.clone_ref(py),
+            Self::Module(module) => module.borrow(py).context().clone_ref(py),
+        }
+    }
+
+    pub(crate) fn merge(
+        py: Python,
+        owners: impl IntoIterator<Item = impl Borrow<Self>>,
+    ) -> PyResult<Self> {
+        owners
```
idavis, line 137: I think we need a code comment on this merge impl

samarsha: I added a doc comment.

idavis (approved):

PR #206, opened 2022-11-29
Add upgrade guide for v0.7

### LINKED ISSUE #203: Migration guide for 0.7
We should add a migration guide to the documentation for how to adapt to the breaking changes in 0.7.

samarsha:
Closes #203.


docs/upgrade/0.7.md
```diff
@@ -0,0 +1,122 @@
...
+   For example:
+
+   ```python
+   from pyqir import Context, SimpleModule, Type
+   context = Context()
+   module = SimpleModule("name", num_qubits=1, num_results=1, context=context)
+   void = Type.void(context)
+   ```
+
+## Parser
```
swernli, line 81: I think it would be helpful to have one more section at the end of this that points to the tests in https://github.com/qir-alliance/pyqir/blob/samarsha/upgrade-guide/pyqir/tests/test_parser.py as an example of how the parser is used so folks can see all these patterns in action. Eventually, we should add two new examples to the examples folder: pure parsing and parse, modify, emit. The new examples can be in a follow up PR.

docs/upgrade/0.7.md
```diff
@@ -0,0 +1,122 @@
...
+
+### Instructions
+
+The instruction class hierarchy was trimmed down significantly.
+Most subclasses of `QirInstr` were removed.
+The surviving subclasses are `Call`, `FCmp`, `ICmp`, `Phi` and `Switch`.
+
+For any other instruction, it should be possible to use the base `Instruction` class to do anything that was previously possible.
+Use the `opcode` property to check what kind of instruction it is, and the `operands` property to read all of the values that the instruction references.
+The `successors` property is a subset of `operands` and contains just the values that are basic blocks, which can be useful to follow control flow with `br` instructions.
```
swernli, line 111: Using the parser to follow control flow is one of the more complex parts for folks using the parser. If a full example won't fit here, could you point to a specific part of the test folks can draw from, such as: https://github.com/qir-alliance/pyqir/blob/a9d6572165496a7ee02e59ffb0b081f0aa10c9a0/pyqir/tests/test_parser.py#L43-L54 and: https://github.com/qir-alliance/pyqir/blob/a9d6572165496a7ee02e59ffb0b081f0aa10c9a0/pyqir/tests/test_parser.py#L64 to show how to go from a terminator with successors to the corresponding block?

swernli (commented):
Some suggestions below on using the tests as examples for the parser changes.

docs/upgrade/0.7.md
samarsha: To go from a terminator with successors to the corresponding block is just `Instruction.successors[i]`. (The type of `successors` is `List[BasicBlock]`.) I'm not sure what example you had in mind? The test snippets you posted are testing block names which isn't necessary for CFG traversal.

swernli: Ah, I see where the text says that, but it wasn't immediately clear to me how easy that would be to use. Maybe just an code snippet that shows grabbing a successor and then investigating its terminator would make that more obvious?

Also, I do really like the design here!

samarsha: I added a sentence with an example.

docs/upgrade/0.7.md
samarsha: I added a section with links to the tests.

idavis (approved):

swernli (approved):

PR #209, opened 2022-12-01
Rewrite SimpleModule in pure Python

samarsha:


pyqir/pyqir/_simple.py
```diff
@@ -0,0 +1,116 @@
...
+    Module,
+    Value,
+)
+
+
+class SimpleModule:
+    """
+    A simple module represents an executable program with these restrictions:
+
+    - There is one global qubit register and one global result register. Both
```
idavis, line 24: Do we want to call these global? Same with the properties below.

idavis (approved):

PR #211, opened 2022-12-02
Generalize API for attributes

samarsha:
I want to make room for non-string-valued attributes as well as parameter and return type attributes. I don't think the current API would be able to do that without breakage, so I am making the breaking changes now:

- Rename `Attribute.value` to `Attribute.string_value` so that non-string value properties can be added later.
- Changed `Function.attribute(kind)` to `Function.attributes.func[kind]`. Other groups are `attributes.param(n)` and `attributes.ret`. The type for these groups acts like a dictionary, with `__contains__` and `__getitem__`. `__setitem__` and `__delitem__` can be implemented later.


pyqir/pyqir/_native.pyi
```diff
@@ -21,8 +21,48 @@ class Attribute:
...
+
+    def __contains__(self, item: str) -> bool:
+        """
+        Tests if an attribute is a member of the group.
+
+        :param item: The attribute kind.
+        :returns: True if the group has an attribute with the given kind.
+        """
+        ...
+    def __getitem__(self, key: str) -> Attribute:
```
idavis, line 39: Doc comment refers to attribute kind instead of name/key

idavis, line 39: Do we need to document that this throws a key error if the key isn't found?

pyqir/tests/test_parser.py
```diff
@@ -186,9 +186,11 @@ def test_parser_internals() -> None:
     assert interop_funcs[0].name == func_name
     assert required_num_qubits(interop_funcs[0]) == 6
 
-    attribute = interop_funcs[0].attribute("requiredQubits")
-    assert attribute is not None
-    assert attribute.value == "6"
+    attributes = interop_funcs[0].attributes.func
```
idavis, line 189: We are missing tests for parameters and return values.

pyqir/src/values.rs
```diff
@@ -462,6 +456,104 @@ impl Function {
...
+/// A group of attributes for a specific part of a function.
+#[pyclass]
+pub(crate) struct AttributeDict {
+    function: Py<Function>,
+    index: AttributeLoc,
+}
+
+#[pymethods]
+impl AttributeDict {
+    /// Tests if an attribute is a member of the group.
```
idavis, line 533: Same as py doc - it is reffering to kind instead of key

pyqir/pyqir/_native.pyi
samarsha: It's intentional. That's LLVM terminology.

samarsha: No exceptions are documented currently.

pyqir/tests/test_parser.py
samarsha: I added some.

idavis (approved):

PR #221, opened 2022-12-20
Adding QIS and RT functions

idavis:
Adding calls:
- barrier
- ccx
- swap
- array_record_output
- initialize
- result_record_output
- tuple_record_output

Moving BasicQisBuilder to Python


pyqir/src/python.rs
```diff
@@ -71,5 +73,32 @@ fn _native(_py: Python, m: &PyModule) -> PyResult<()> {
...
+
+    m.add_wrapped(wrap_pymodule!(_qis))?;
+    let sys = PyModule::import(py, "sys")?;
+    let sys_modules: &PyDict = sys.getattr("modules")?.downcast()?;
+    sys_modules.set_item("pyqir.qis._native", m.getattr("_qis")?)?;
+
+    m.add_wrapped(wrap_pymodule!(_rt))?;
+    let sys = PyModule::import(py, "sys")?;
+    let sys_modules: &PyDict = sys.getattr("modules")?.downcast()?;
+    sys_modules.set_item("pyqir.rt._native", m.getattr("_rt")?)?;
```
samarsha, line 85: This is really hacky... It'd be easier not to deal with the submodules in pyo3, you can add everything to pyqir._native, but re-export them from qis.py and rt.py instead of \_\_init\_\_.py.

pyqir/tests/test_qis.py
```diff
@@ -31,6 +32,39 @@ def test_single(
...
+    name: str,
+    get_gate: Callable[[BasicQisBuilder], Callable[[Value, Value, Value], None]],
+) -> None:
+    mod = SimpleModule("test_three_qubit_gates", 3, 0)
+    basic = BasicQisBuilder(mod.builder)
+    get_gate(basic)(mod.qubits[0], mod.qubits[1], mod.qubits[2])
+    call = f"call void @__quantum__qis__{name}__body(%Qubit* null, %Qubit* inttoptr (i64 1 to %Qubit*), %Qubit* inttoptr (i64 2 to %Qubit*))"
+    assert call in mod.ir()
+
+
```
samarsha, line 67: barrier?

pyqir/src/rt.rs
```diff
@@ -0,0 +1,94 @@
...
+};
+use pyo3::prelude::*;
+use qirlib::rt;
+
+/// Inserts a marker in the generated output that indicates the start
+/// of an array and how many array elements it has.
+///
+/// :param Builder builder: The IR Builder used to create the instructions
+/// :param Value num_elements: How many array elements the array has
+/// :param str label: A string label for the array. Depending on the output schema, the label is included in the output or omitted.
```
samarsha, line 16:
```suggestion
/// :param Value label: A string label for the array. Depending on the output schema, the label is included in the output or omitted.
```

pyqir/src/rt.rs
```diff
@@ -0,0 +1,94 @@
...
+        rt::build_array_record_output(builder.as_ptr(), num_elements.as_ptr(), label.as_ptr());
+    }
+    Ok(())
+}
+
+/// Adds a measurement result to the generated output.
+///
+/// :param Builder builder: The IR Builder used to create the instructions
+/// :param Value result: A result measurement to record
+/// :param str label: A string label for the result value. Depending on the output schema, the label is included in the output or omitted.
```
samarsha, line 38:
```suggestion
/// :param Value label: A string label for the result value. Depending on the output schema, the label is included in the output or omitted.
```

pyqir/src/rt.rs
```diff
@@ -0,0 +1,94 @@
...
+    }
+    Ok(())
+}
+
+/// Inserts a marker in the generated output that indicates the start
+/// of a tuple and how many tuple elements it has.
+///
+/// :param Builder builder: The IR Builder used to create the instructions
+/// :param Value num_elements: How many tuple elements the tuple has
+/// :param str label: A string label for the tuple. Depending on the output schema, the label is included in the output or omitted.
```
samarsha, line 78:
```suggestion
/// :param Value label: A string label for the tuple. Depending on the output schema, the label is included in the output or omitted.
```

pyqir/src/rt.rs
```diff
@@ -0,0 +1,94 @@
...
+
+/// Initializes the execution environment. Sets all qubits to a zero-state
+/// if they are not dynamically managed.
+///
+/// :param Builder builder: The IR Builder used to create the instructions
+/// :param Value reserved: Reserved. For base profile QIR, a const null i8* Value should be passed.
+#[pyfunction]
+#[pyo3(text_signature = "(builder, reserved)")]
+#[allow(clippy::needless_pass_by_value)]
+pub(crate) fn initialize(py: Python, builder: Py<Builder>, reserved: &Value) -> PyResult<()> {
```
samarsha, line 64: Maybe something generic like `data` instead of `reserved`? (Just thinking since changing param name in the future can be breaking.)

pyqir/src/qis.rs
```diff
@@ -21,13 +21,77 @@ pub(crate) struct BasicQisBuilder {
...
+            qis::build_ccx(
+                builder.as_ptr(),
+                control1.as_ptr(),
+                control2.as_ptr(),
+                target.as_ptr(),
+            );
+        }
+        Ok(())
+    }
+
```
samarsha, line 94: I like the idea of the qis and rt functions being standalone, but the split for qis right now seems awkward. Why is ccx is BasicQisBuilder but barrier and swap in qis? For 0.8, should we copy all BasicQisBuilder methods to qis, then convert it to  a Python wrapper class like we talked about?

samarsha (changes requested):

pyqir/pyqir/_basicqis.py
```diff
@@ -0,0 +1,174 @@
...
+    def ccx(self, control1: Value, control2: Value, target: Value) -> None:
+        """
+        Inserts Toffoli or doubly-controlled :math:`X` gate.
+
+        :param Value control1: The first control qubit.
+        :param Value control2: The second control qubit.
+        :param Value target: The target qubit.
+        :rtype: None
+        """
+        qis.ccx(self._builder, control1, control2, target)
```
samarsha, line 29: Can we remove this since it isn't needed for BC?

pyqir/pyqir/_native.pyi
```diff
@@ -986,3 +846,242 @@ def result_type(context: Context) -> Type:
...
+# Runtime
+
+def array_record_output(builder: Builder, num_elements: Value, label: Value) -> None:
+    """
+    Inserts a marker in the generated output that indicates the start
+    of an array and how many array elements it has.
+
+    :param Builder builder: The IR Builder used to create the instructions
+    :param Value num_elements: How many array elements the array has
+    :param str label: A string label for the array. Depending on the output schema, the label is included in the output or omitted.
```
samarsha, line 859: Should be type Value, not str. But you don't need types in the docstrings since this file has type annotations, so you could remove them here.

pyqir/src/qis.rs
```diff
@@ -12,309 +12,408 @@ use llvm_sys::{
...
 ///
-/// :param Builder builder: The underlying builder used to build QIS instructions.
-#[pyclass]
+/// :param builder: The underlying builder used to build QIS instructions.
+/// :rtype: None
+#[pyfunction]
 #[pyo3(text_signature = "(builder)")]
-pub(crate) struct BasicQisBuilder {
+#[allow(clippy::needless_pass_by_value)]
+pub(crate) fn barrier(py: Python, builder: Py<Builder>) {
```
samarsha, line 22: For all of these, you should be able to take in a `builder: &Builder`, which saves you needing to borrow it explicitly.

pyqir/src/qis.rs
```diff
@@ -12,309 +12,408 @@ use llvm_sys::{
 use pyo3::prelude::*;
 use qirlib::qis;
 
-/// An instruction builder that generates instructions from the basic quantum instruction set.
+/// Inserts a barrier instruction
 ///
-/// :param Builder builder: The underlying builder used to build QIS instructions.
-#[pyclass]
+/// :param builder: The underlying builder used to build QIS instructions.
```
samarsha, line 17: "Underlying builder" sounds odd since it's not being stored in a class anymore, can you use the same wording as the rt functions?

qirlib/src/utils.rs
```diff
@@ -0,0 +1,126 @@
...
+pub(crate) unsafe fn function_type(ret: LLVMTypeRef, params: &mut [LLVMTypeRef]) -> LLVMTypeRef {
+    LLVMFunctionType(
+        ret,
+        params.as_mut_ptr(),
+        params.len().try_into().unwrap(),
+        0,
+    )
+}
+
+pub(crate) unsafe fn declare(
```
samarsha, line 97: What about declare_qis and declare, instead of declare and declare_bare?

qirlib/src/utils.rs
```diff
@@ -0,0 +1,127 @@
...
+    declare_external_function(module, name.as_str(), ty)
+}
+
+pub(crate) unsafe fn declare_external_function(
+    module: LLVMModuleRef,
+    name: &str,
+    ty: LLVMTypeRef,
+) -> LLVMValueRef {
+    let name = CString::new(name)
+        .expect("Could not create QIS declaration from name containing a null byte");
```
samarsha, line 118:
```suggestion
        .expect("Could not create declaration from name containing a null byte");
```

samarsha (approved):

PR #222, opened 2022-12-20
Changing build to target LLVM 14

idavis:


samarsha (commented):
About the deprecated functions, it looks like the replacements LLVMBuildCall2 and LLVMConstGEP2 are available back to at least LLVM 11. Do they work here or do they require opaque pointers?

samarsha (approved):

PR #225, opened 2022-12-22
LLVM Build Caching

idavis:


.github/actions/install-llvm/action.yml
```diff
@@ -0,0 +1,66 @@
...
+    - name: Windows - Install build dependencies, ninja
+      run: choco install --accept-license -y ninja
+      shell: pwsh
+      if: ${{ (inputs.os == 'windows-2019') && (steps.cache-llvm.outputs.cache-hit != 'true') }}
+    - name: MacOS - Install build dependencies, ninja
+      run: brew install ninja
+      shell: pwsh
+      if: ${{ (inputs.os == 'macos-11') && (steps.cache-llvm.outputs.cache-hit != 'true') }}
+    - name: Windows - Install LLVM 14.0.6
+      run: choco install llvm --version=14.0.6 --allow-downgrade
```
samarsha, line 45: Why does this ignore inputs.version?

rust-toolchain.toml
```diff
@@ -0,0 +1,3 @@
+[toolchain]
+channel = "1.64.0"
+components = [ "rustfmt", "clippy" ]
```
samarsha, line 3: Is this necessary? I'd rather be able to use the latest toolchain locally. Also, doesn't this duplicate the Rust version in psakefile.ps1 and Dockerfile.manylinux?

.github/actions/install-llvm/action.yml
```diff
@@ -0,0 +1,66 @@
...
+    description: "The OS being built upon."
+    required: true
+  directory:
+    description: "The directory to install LLVM binaries to."
+    required: true
+  target:
+    description: "The build script target."
+    required: true
+  arch:
+    description: "TODO: The OS arch being built for."
```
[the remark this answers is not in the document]
idavis: Placeholder to put in for arm or x64, but we can kill it for now.

.github/actions/install-llvm/action.yml
idavis: That is the clang tooling used to build LLVM, not the LLVM version to build

.github/actions/install-llvm/action.yml
```diff
@@ -0,0 +1,66 @@
...
+
+runs:
+  using: composite
+  steps:
+    - name: LLVM install cache
+      id: cache-llvm
+      uses: actions/cache@v3
+      with:
+        path: ${{ inputs.directory }}
+        key: llvm-${{ inputs.version }}-${{ inputs.arch }}-${{ inputs.os }}-${{ inputs.target }}-${{ hashFiles('Cargo.lock') }}
```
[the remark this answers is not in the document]
idavis: If we update our build deps, we want the cache invalidated, llvm-sys, cc, cmake, lazy_static

rust-toolchain.toml
idavis: You can still use the latest with the toolchain file there. It makes loading the correct rust toolchain very simple for CI. If you set the rustup override you'll continue to use your global override. If we move the the toolchain file we can remove the other instances you mention and load data from the toolchain file.

.github/actions/install-llvm/action.yml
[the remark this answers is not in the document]
samarsha: But this cache only stores the build artifacts for LLVM itself. That shouldn't depend on the Rust crates we use?

rust-toolchain.toml
samarsha: > You can still use the latest with the toolchain file there.

It doesn't seem like it... when I try this branch, cargo will automatically download Rust 1.64 and use it, even though I have 1.66 installed. That's what I'd like to avoid.

.github/actions/install-llvm/action.yml
samarsha: I think it'd be less confusing if this was in the earlier "Windows - Install build dependencies" step.

eng/psakefile.ps1
```diff
@@ -30,15 +29,19 @@ task run-manylinux-container-image -preaction { Write-CacheStats } -postaction {
     # docker will create it and it will be owned by root and
     # the caching/install breaks with permission errors.
     # New-Item is idempotent so we don't need to check for existence
-    $cacheMount, $cacheEnv = Get-CCacheParams
+    $llvmDir = Resolve-InstallationDirectory
+    if (!(Test-Path $llvmDir)) {
+        New-Item -ItemType Directory -Force $llvmDir | Out-Null
+    }
```
samarsha, line 35: This is used in 3 places, would be nice to have a function that both creates and returns an installation directory.

.github/actions/install-llvm/action.yml
idavis: Did some retesting and llvm on windows is no longer needed in the builds. I now uninstall it from the images.

samarsha (approved):

PR #230, opened 2023-01-22
Add ccx gate

LaurentAjdnik:
CCX gate declared elsewhere but not in `_basicqis.py`, resulting in `AttributeError: 'builtins.BasicQisBuilder' object has no attribute 'ccx'`.

### DIFF
--- a/pyqir/pyqir/_basicqis.py
+++ b/pyqir/pyqir/_basicqis.py
@@ -26,6 +26,16 @@ class BasicQisBuilder:
         """
         qis.cx(self._builder, control, target)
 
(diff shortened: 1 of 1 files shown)


samarsha (changes requested):
Thanks for your PR. We left `ccx` out of `BasicQisBuilder` as part of a transition to a newer `pyqir.qis.*` and `pyqir.rt.*` API. So our intention is for users to call `pyqir.qis.ccx` directly, and `BasicQisBuilder` may be deprecated in the future (although we haven't decided exactly what to do about it).

PR #232, opened 2023-04-13
Adding LLVM value equality and hash implementation.

idavis:


swernli (approved):

pyqir/tests/test_value_operators.py
```diff
@@ -0,0 +1,175 @@
...
+
+
+def test_instruction_hash_equals_self_hash() -> None:
+    inst = get_instructions()[0]
+    assert hash(inst) == hash(inst)
+
+
+def test_instruction_hash_not_equals_other_hash() -> None:
+    insts = get_instructions()
+    assert hash(insts[0]) != hash(insts[1])
```
samarsha, line 78: I'm not sure I would test this. It's not a guaranteed property based on the interface, even though it's probably the case that the hasher just copies the pointer into the hash on 64-bit systems.

idavis: You're right.

Issue #250, opened 2023-09-18
Misnamed Function Attributes (required_num_* vs num_required_*)


bmhowe23:
**Describe the bug**

The [Base Profile Spec](https://github.com/qir-alliance/qir-spec/blob/main/specification/under_development/profiles/Base_Profile.md) specifies that the `required_num_qubits` and `required_num_results` attributes must be attached to the entry point function, and while pyqir uses those attribute names in some places (like for internal variable names), there are some important places (like the LLVM interface) where `num_required_qubits` and `num_required_results` are used instead.

**To Reproduce**

1. Modify `qirlib/resources/tests/rt/initialize.ll` to match the Base Profile specification by applying this patch:
```diff
diff --git a/qirlib/resources/tests/rt/initialize.ll b/qirlib/resources/tests/rt/initialize.ll
index 5a88b41..9ba9b22 100644
--- a/qirlib/resources/tests/rt/initialize.ll
+++ b/qirlib/resources/tests/rt/initialize.ll
@@ -8,4 +8,4 @@ define void @main() #0 {

 declare void @__quantum__rt__initialize(i8*)

-attributes #0 = { "entry_point" "num_required_qubits"="0" "num_required_results"="0" "output_labeling_schema" "qir_profiles"="custom" }
+attributes #0 = { "entry_point" "output_labeling_schema" "qir_profiles"="custom" "required_num_qubits"="0" "required_num_results"="0" }
```
2. Run `build.ps1`
3. Look for this failure message:
```bash
test rt::tests::initialize ... FAILED
```

**Expected behavior**

The test should not fail when using spec-compliant attributes.

**Screenshots**

N/A

**System information**

- Release version or commit for which the problem occurs: de2c8ed (current latest)
- Your operating system and Python or Rust version: Ubuntu 22.04, rustc 1.72.0 (5680fa18f 2023-08-23)

**Additional context**

N/A

idavis:
Hi @bmhowe23 - I need to look at this. To my knowledge the base profile is only defined in [qir-spec #25 PR](https://github.com/qir-alliance/qir-spec/pull/25) which I just saw you had some comments on. Everything here is based on that PR trying to define the base profile.

idavis:
Fixed in #251, not yet released. Thanks @bmhowe23 !

PR #251, opened 2023-09-18
Spec compliance: change num_required_* to required_num_*

bmhowe23:
As described in Issue #250, the Base Profile spec specifies that `required_num_qubits` and `required_num_results` must be used in the entry point function. This change updates the pyqir project to use those attributes in all places. I _think_ it also maintains ability to parse old `num_required_*` attributes in order to maintain as much backwards compatibility as possible, but please let me know if I've missed anything there.


bmhowe23:
@microsoft-github-policy-service agree [company="NVIDIA"]

bmhowe23:
@microsoft-github-policy-service agree company="NVIDIA"

PR #275, opened 2024-05-22
Removing universal build, make cross explicit

idavis:


idavis:
This was a test to see what the CI builds would do.

Issue #224, opened 2022-12-20
Add LLVM Opaque Pointer Support


idavis:
Once [Opaque Pointers in QIR](https://github.com/qir-alliance/qir-spec/issues/30) is resolved, the deprecated calls need to be replaced with the new versions:
| Old Name | New Name |
|----------|----------|
|LLVMBuildLoad | LLVMBuildLoad2 |
|LLVMBuildCall | LLVMBuildCall2 |
|LLVMBuildInvoke | LLVMBuildInvoke2 |
|LLVMBuildGEP | LLVMBuildGEP2 |
|LLVMBuildInBoundsGEP | LLVMBuildInBoundsGEP2 |
|LLVMBuildStructGEP | LLVMBuildStructGEP2 |
|LLVMBuildPtrDiff | LLVMBuildPtrDiff2 |
|LLVMConstGEP | LLVMConstGEP2 |
|LLVMConstInBoundsGEP | LLVMConstInBoundsGEP2 |
|LLVMAddAlias | LLVMAddAlias2 |

PR #278, opened 2024-06-03
Get building on LLVM 18.1 (...But not working)

ausbin:
Rough work on upgrading to LLVM 18.1.6 (#224). This builds, but the tests definitely do not pass (some segfault, actually). So this should **not** be merged into `main` in its current state, but I'm putting this up if it saves someone some time.

I built this with the following in `pyqir/`:
```
$ pip install -v --config-settings=build-args='-F qirlib/llvm18-1 -F pyqir/llvm18-1' .[test]
```
(I did not use the PowerShell build script.)

### Remaining Problems
If y'all are interested, I can continue working on this and get the tests to pass.

One thing that seems particularly nasty at the moment is how to handle `Builder.call()`. I can think of two options:
1. Modify the signature of `Builder.call()` to also take a `Type` (looks like that already exists in `types.rs`). This is backwards incompatible
2. Modify `Value` to also hold an `LLVMType`. (This may quickly become burdensome to support in future usecases, but it would probably be a simpler user-facing API)

What do y'all think? And are there plans to upgrade LLVM at all?

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -12,6 +12,12 @@ dependencies = [
 ]
 
 [[package]]
+name = "anyhow"
+version = "1.0.86"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "b3d1d046238990b9cf5bcde22a3fb3584ee5cf65fb2765f454ed428c7a0063da"
+
+[[package]]
 name = "autocfg"
 version = "1.1.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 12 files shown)


ausbin:
Just stumbled upon this: https://github.com/qir-alliance/pyqir/blob/c02430a5d4f566425f6112576d4109071011ea66/qirlib/src/types.rs#L43. That will [need][1] to get fixed too

[1]: https://llvm.org/docs/OpaquePointers.html

ausbin:
I'm getting the vibe that it is not helpful for me to continue working on this, closing

Issue #277, opened 2024-05-28
Support for loops


cabreraam:
**Is your feature request related to a problem? Please describe.**
When a `for` loop in PyQIR is encountered, the resulting QIR that's emitted is effectively the loop unrolled equivalent of that `for` loop. Is there a way to use the branching infrastructure in LLVM to reduce the size of the resulting `.ll` file? I'm not sure if this is _physically_ realizable on a QPU (I'm very new to this space), so forgive me if this is a naive question.

**Describe the solution you had in mind**
It seems like most of the component pieces for a solution (in software) exist:
* you have the example [`if_bool.py`](https://github.com/qir-alliance/pyqir/blob/main/examples/if_bool.py) example for executing different basic blocks based on some condition
* [a stub for implementing a `phi` node](https://github.com/qir-alliance/pyqir/blob/c02430a5d4f566425f6112576d4109071011ea66/pyqir/pyqir/_native.pyi#L739-L745)

I don't know anything about Rust, so I don't understand the function signatures for implementations, but it feels like the infrastructure for a solution is there. 

**Additional context**
Maybe @wongey or @ausbin might be able to provide additional context/correct some things that I wrote.

cabreraam:
This bit of LLVM IR doesn't do anything useful, but this is something I rolled by hand and is what I had in mind. Also, it works in [this build of `qir-runner`](https://github.com/qir-alliance/qir-runner/actions/runs/9275389338)

```llvm
; ModuleID = 'qpe'
source_filename = "qpe"

%Qubit = type opaque

define void @main() #0 {
entry:
  br label %loop0

loop0:
  %i = phi i64 [ 0, %entry ], [ %nextval0, %loop0 ]
  %tmp = inttoptr i64 %i to %Qubit*
  call void @__quantum__qis__h__body(%Qubit* %tmp)
  %nextval0 = add i64 %i, 1
  %endcond0 = icmp ult i64 %i, 3
  %loopcond0 = icmp ne i1 %endcond0, 0
  br i1 %loopcond0, label %loop0, label %afterloop0

afterloop0:
  ret void
}


declare void @__quantum__qis__h__body(%Qubit*)

attributes #0 = { "entry_point" "output_labeling_schema" "qir_profiles"="custom" "required_num_qubits"="4" "required_num_results"="3" }

!llvm.module.flags = !{!0, !1, !2, !3}

!0 = !{i32 1, !"qir_major_version", i32 1}
[...]

Issue #242, opened 2023-07-10
Add support for phi node generation


idavis:
A support for phi nodes in QIR generation API

idavis:
Fixed by #279

Issue #228, opened 2023-01-12
Problem and Solution for Consolidating Build Setup - Suggested contribution


owen-oqc:
**Is your feature request related to a problem? Please describe.**

The build instructions differ significantly from OS to OS and require a number of automated installation steps.

**Describe the solution you had in mind**

I have worked through your build instructions and generated a conda env that pulls all the requirements together. That would make the setup steps looks like.

```bash
conda env create -f pyqir.yaml && conda activate pyqir
```

Which can bring in cmake, rust, python 3.9, ....

This works well on my OSX M1 env 

**Additional context**

As I have done this i'd gladly make the contribution directly if you wanted it.

ausbin:
If anyone else is not interested in installing PowerShell, it seems you can build and test the Python extension as follows (assuming `llvm-config` for your LLVM build is in your `$PATH`):
```
$ python3 -m venv venv
$ . venv/bin/activate
$ cd pyqir
$ pip install -v --config-settings=build-args='-F qirlib/llvm14-0 -F pyqir/llvm14-0' .[test]
$ pytest
```
Works for me on Debian bookworm

PR #279, opened 2024-06-04
Add API for inserting phi nodes

ausbin:
Allows inserting phi nodes and conditional branches. Hopefully addresses #242 and possibly #277

### Testing
`pytest` passes with a local build of LLVM 14.0.6. I am building with the incantation [here][1]

I added a low-effort unit test. If y'all are interested in merging I can put more effort into adding new unit tests (when the giga-test I added fails, the output from pytest is pretty much unreadable)

[1]: https://github.com/qir-alliance/pyqir/issues/228#issuecomment-2146277863


pyqir/src/builder.rs
```diff
@@ -199,6 +201,21 @@ impl Builder {
         }
     }
 
+    /// Converts a dynamically-calculated qubit index to a Qubit pointer
+    ///
+    /// :param Value id: The qubit id as a Value
+    /// :returns: A Value for a Qubit pointer
+    /// :rtype: Value
+    #[pyo3(text_signature = "(self, id)")]
+    fn dyn_qubit(&self, py: Python, id: &Value) -> PyResult<PyObject> {
```
ausbin, line 210: Not the most beautiful name or place for this, but I needed it and couldn't find it anywhere. All I wanted was to create an [`inttoptr`][1] with an input SSA edge but `pyqir.qubit()` seems to expect a constant `int`

[1]: https://llvm.org/docs/LangRef.html#inttoptr-to-instruction

idavis:
To address this, I have updated #224 and created #280 as a tracking issue. A high level goal is to enable typed/opaque pointers using LLVM 16 to allow migration from the older typed pointers in QIR and then future versions would all be using the opaque APIs.

ausbin:
@idavis I'm a little confused by your response, is that a note to self or something actionable on my end?

Also, is there a definition of the adaptive profile online? I only know about [this Q# documentation][1], but I was hoping for something analogous to [the description of the Base Profile][2].

[1]: https://learn.microsoft.com/en-us/azure/quantum/quantum-computing-target-profiles
[2]: https://github.com/qir-alliance/qir-spec/blob/main/specification/under_development/profiles/Base_Profile.md

idavis:
@ausbin Could you please separate the phi node support from the dyn qubits? I like the phi node work. I'm hesitant to add the dyn qubits as the QIR profiles haven't defined what dynamic qubits will look like.

Here is the current [Adaptive Profile](https://github.com/qir-alliance/qir-spec/blob/11db5f743d637faef6b6d4ebab3ae7c8d6fe87c5/specification/under_development/profiles/Adaptive_Profile.md). Using `phi` nodes requires the integer computation extension.

pyqir/pyqir/_simple.py
```diff
@@ -84,6 +85,16 @@ def builder(self) -> Builder:
...
+    @property
+    def entry_point(self) -> Function:
+        """The entry point function (automatically generated)."""
+        return self._entry_point
+
+    @property
+    def entry_block(self) -> BasicBlock:
+        """The first basic block of the entry point (automatically generated)."""
+        return self._entry_block
+
```
ausbin, line 97: This is used in the test I added, but it should be pretty useful to users too if they are manually messing with basic blocks

ausbin:
Thanks for the link, @idavis. I will take a look.

I removed the `dyn_qubit()` stuff and added a couple of tests, is that better?

idavis (approved):

PR #298, opened 2024-10-15
Add ability to set name on values

idavis:


pyqir/tests/test_value_name.py
```diff
@@ -0,0 +1,73 @@
...
+    sink_res = mod.builder.call(sink, [source_res])
+    sink_res.name = "my_res"
+    ir = mod.ir()
+    assert "%my_var = call i64 @source()" in ir
+    assert "%my_res = call i64 @sink(i64 %my_var)" in ir
+
+
+def test_function_name_can_contain_spaces_and_chars() -> None:
+    mod = pyqir.SimpleModule("test", 0, 0)
+    expected = "Some - ; name fin"
```
swernli, line 67: It's surprising to me that this works... out of curiosity, does the module succeed `verify` even with this?

swernli (approved):

pyqir/tests/test_value_name.py
idavis: Added it to the test.

PR #306, opened 2025-01-23
Upgrade pyo3 from v0.20 to v0.22

sezna:
This is currently a work in progress, but from a high level, this PR:
- Uses the new `Bound` API
- Replaces deprecated implicit "option=None" func args, making those explicit (the changes from text_signature to signature)
- Updates pyclass to the newly required `pyclass(eq, eq_int)` at _least_, and include `ord` if we had manually implemented `__richcmp__` before
- Other small changes according to the migration guide

Used my learnings from when I [upgraded qsharp's pyo3 version](https://github.com/microsoft/qsharp/pull/1893) and the regression I caused, which @minestarks [later fixed](https://github.com/microsoft/qsharp/pull/1979/files).


swernli:
Superseded by #314

PR #307, opened 2025-01-23
update rust toolchain from 1.80 to 1.82

sezna:
This rust version matches what we use in [microsoft/qsharp](github.com/microsoft/qsharp).


swernli:
We already updated to 1.84... we should probably do that here too, which changes which pyO3 we need to pull in too...
https://github.com/microsoft/qsharp/pull/2107

sezna:
Hm, ok, I'm going to close this one until we get to pyo3 0.23 over here, then.

Issue #308, opened 2025-01-30
Add Support for manylinux 2 28


cqc-melf:
We are currently running issues using the pyqir wheel for linux you provide on pypi because of dependency issues.
The version you provide requires glibc 2 31, we would like to use glibc 2.28.

Would it be possible for you to change the audit tag from to `manylinux_2_28_x86_64` (from currently `manylinux_2_31_x86_64` )

I have set up a PR which I hope would you introduce this change. You can find it at: ( https://github.com/qir-alliance/pyqir/pull/309 )

If you have any questions, please let me know.

idavis:
As mentioned in #309, 2.28 support requires a linux version that is seven years old and passed it EOL. Only 2.31 and above are available in supported base images and we don't have infrastructure to support it.

ElieGouzien:
@cqc-melf : if you ended here because of using pixi, make sure to check this documentation:
https://pixi.prefix.dev/dev/workspace/system_requirements/

In particular, `libc = { family = "glibc", version = "2.41" }` works.

PR #309, opened 2025-01-30
Update psakefile.ps1, change to manylinux_2_28_x86_64

cqc-melf:
This PR tries to solve https://github.com/qir-alliance/pyqir/issues/308


idavis:
Unfortunately supporting `manylinux_2_28_x86_64` would require building on images that aren't available in the production pipeline.

PR #314, opened 2025-02-22
Update to pyO3 v0.23

swernli:
This picks up where #306 left off and completes the next phase to update to pyO3 v0.23.


orpuente-MS:
Looks good. Approved with minor comment.

pyqir/src/module.rs
```diff
@@ -64,14 +64,18 @@ impl Module {
...
     #[staticmethod]
-    #[pyo3(text_signature = "(context, ir, name=\"\")")]
+    #[pyo3(signature = (context, ir, name=""))]
     fn from_ir(py: Python, context: Py<Context>, ir: &str, name: Option<&str>) -> PyResult<Self> {
         let name = CString::new(name.unwrap_or_default()).unwrap();
 
+        // LLVMParseIRInContext takes a null-terminated string, so use a
+        // CString to ensure safety across the FFI boundary.
+        let len = ir.len();
+        let ir = CString::new(ir).unwrap();
```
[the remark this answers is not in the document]
swernli: I updated the patterns in this function, but generally throughout the PyQIR code we used `unwrap` instead of `expect` so I was matching style. It might be worth doing an update and enabling the clippy error for unwrap usage like we did in the qsharp repo.

pyqir/src/module.rs
```diff
@@ -64,27 +64,37 @@ impl Module {
...
+    #[pyo3(signature = (context, ir, name=""))]
     fn from_ir(py: Python, context: Py<Context>, ir: &str, name: Option<&str>) -> PyResult<Self> {
-        let name = CString::new(name.unwrap_or_default()).unwrap();
+        let name = CString::new(name.unwrap_or_default())
+            .expect("should be able to create CString for name");
+
+        // LLVMParseIRInContext takes a null-terminated string, so use a
+        // CString to ensure safety across the FFI boundary.
+        let len = ir.len();
+        let ir = CString::new(ir).expect("should be able to create CString for ir");
```
minestarks, line 75: Was this a bug before?

minestarks (approved):

pyqir/src/module.rs
swernli: In a very subtle way, yes! When I updated from PyO3 0.22 to 0.23, I started seeing memory corruptions and random test failures. After much investigation and with help from @orpuente-MS, we narrowed it down to this input string. Fascinatingly, if we used a hard coded string rather than the one that was passed in from Python, it failed on both version. Null termination ended up being the culprit...

In Rust, &str slices are defined as a pair of bytes and length. But since the length is included in the type they are NOT guaranteed to be null terminated. Meanwhile, a CString specifically calls out that it is null terminated for compatibility across FFI:

> A type representing an owned, C-compatible, nul-terminated string with no nul bytes in the middle.
> This type serves the purpose of being able to safely generate a C-compatible string from a Rust byte slice or vector. An instance of this type is a static guarantee that the underlying bytes contain no interior 0 bytes ("nul characters") and that the final byte is 0 ("nul terminator").

I expect that somewhere in PyO3 it used to do something extra to ensure the strings were null-terminated, and now it more "correctly" does FFI to Rust by creating string slices that have length and don't have the null-terminator... which is fine if you stay in Rust and poisonous if that string is then passed down into C.

minestarks: Nice! That must have been painful, but I'm entertained :P

PR #348, opened 2025-12-09
Simplifying build, removing deps

idavis:


swernli (approved):
Good finds, thanks for fixing all this!

PR #358, opened 2026-04-21
Fix segmentation fault when accessing switch instruction cond value

### LINKED ISSUE #357: [SEGFAULT] Segmentation fault when accessing switch instruction cond value
**Describe the bug**

There is a segfault in pyqir when trying to access a switch instruction's cond value.

**To Reproduce**

1. Create a file with the following python code
2. Run the file with `python <FILE_NAME>`
3. See error

```python
import pyqir

SWITCH_CASE_QIR = r"""
%Result = type opaque
%Qubit = type opaque

define i64 @ENTRYPOINT__main() #0 {

entry:
  %val = add i64 0, 1
  switch i64 %val, label %default [
    i64 0, label %case0
    i64 1, label %case1
    i64 2, label %case2
  ]
case0:
  br label %measure
case1:
  ; This is the expected path for val==1
  call void @__quantum__qis__x__body(%Qubit* inttoptr (i64 0 to %Qubit*))
  br label %measure
case2:
  br label %measure
default:
  br label %measure
measure:
  call void @__quantum__qis__mresetz__body(%Qubit* inttoptr (i64 0 to %Qubit*), %Result* inttoptr (i64 0 to %Result*))

  call void @__quantum__rt__tuple_record_output(i64 1, i8* null)
  call void @__quantum__rt__result_record_output(%Result* inttoptr (i64 0 to %Result*), i8* null)
  ret i64 0
}

declare void @__quantum__qis__x__body(%Qubit*)
declare void @__quantum__qis__mresetz__body(%Qubit*, %Result*)
declare void @__quantum__rt__tuple_record_output(i64, i8*)
declare void @__quantum__rt__result_record_output(%Result*, i8*)

attributes #0 = { "entry_point" "qir_profiles"="adaptive_profile" "required_num_qubits"="1" "required_num_results"="1" }
attributes #1 = { "irreversible" }
"""


def main():
    context = pyqir.Context()
    mod = pyqir.Module.from_ir(context, SWITCH_CASE_QIR)
    switch_instr = mod.functions[0].basic_blocks[0].instructions[1]
    assert isinstance(switch_instr, pyqir.Switch)

    print(switch_instr.default) # this works
    print(switch_instr.operands[0])    # this works
    print(switch_instr.cond) # this segfaults


if __name__ == "__main__":
    main()
```

**Expected behavior**
The `cond` value should be printed; but the program segfaults.

**System information**

The segfault is verified to occur on pyqir `v0.11.1` and `v0.12.3` on Windows and macOS, I did not try Linux.

**Workaround**

If you hit this issue you can work around it by doing:

```python
switch_instr.operands[0] # this is the raw location of the cond value
```

swernli:
Fixes #357

### DIFF
--- a/pyqir/src/instructions.rs
+++ b/pyqir/src/instructions.rs
@@ -356,7 +356,7 @@ impl Switch {
     fn cond<'py>(slf: PyRef<Self>, py: Python<'py>) -> PyResult<Bound<'py, PyAny>> {
         let slf = slf.into_super().into_super();
         unsafe {
-            let value = LLVMGetCondition(slf.cast().as_ptr());
+            let value = LLVMGetOperand(slf.cast().as_ptr(), 0);
             Value::from_raw(py, slf.owner().clone_ref(py), value)
         }
     }
--- a/pyqir/tests/switch_instr.ll
+++ b/pyqir/tests/switch_instr.ll
@@ -1,0 +1,37 @@
+%Result = type opaque
+%Qubit = type opaque
+
+define i64 @ENTRYPOINT__main() #0 {
+
+entry:
+  %val = add i64 0, 1
+  switch i64 %val, label %default [
+    i64 0, label %case0
+    i64 1, label %case1
+    i64 2, label %case2
+  ]
+case0:
+  br label %measure
+case1:
+  ; This is the expected path for val==1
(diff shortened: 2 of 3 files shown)


orpuente-MS (approved):
Nice! Thanks.

PR #360, opened 2026-05-04
Add global variables and array value introspection

orpuente-MS:
This PR adds global variables and array value introspection to pyqir. The goal is to support global arrays introspection.

### DIFF
--- a/pyqir/tests/test_global_variables.py
+++ b/pyqir/tests/test_global_variables.py
@@ -187,7 +187,6 @@ def test_multiple_globals_mixed_types() -> None:
 
     gv_array = mod.global_variables[2]
     assert gv_array.name == "an_array"
-
     assert isinstance(gv_array.initializer, ArrayConstant)
 
     gv_ext = mod.global_variables[3]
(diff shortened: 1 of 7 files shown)


pyqir/tests/test_global_variables.py
```diff
@@ -0,0 +1,205 @@
...
+    define void @func() {
+      ret void
+    }
+
+    declare void @decl()
+    """
+    mod = Module.from_ir(Context(), ir, "test")
+    gv_names = [gv.name for gv in mod.global_variables]
+    assert "func" not in gv_names
+    assert "decl" not in gv_names
```
swernli: nitpick: these two checks aren't really needed, since the entire contents of the list is checked below.

pyqir/tests/test_global_variables.py
```diff
@@ -0,0 +1,205 @@
...
+    assert isinstance(gv_array.initializer, ArrayConstant)
+
+    gv_ext = mod.global_variables[3]
+    assert gv_ext.name == "external"
+    assert gv_ext.initializer is None
+    assert gv_ext.is_constant is False
+
+
+def test_global_with_internal_linkage() -> None:
+    ir = '@.str = private constant [6 x i8] c"hello\\00"'
```
swernli, line 200: we do have some other infrastructure for getting constant strings, would be good to double check if that and this new global constant feature play nicely together.

swernli (approved):
a few minor comments, otherwise this looks good!

pyqir/tests/test_global_variables.py
orpuente-MS: Good point. Added a test verifying this.

